PR #19, opened 2013-12-09
(feature) [WebDriver] Some useful features for WebDriver

Cka3o4Huk:
1. Macroses

Here is basic macroses:

go("${host}") - navigate to ${host}
iWait(function(){ ... }) - specify implicit wait as 7 seconds for lookups of elements inside function
_iWait(function(){ ... }, N) - specify implicit wait as N seconds for lookups of elements inside function
$("#myid") - lookup single element by CSS
$$("#myid divs") - lookup list of element by CSS
$x(..) - lookup single element by XPath
measure(function(){ ... }) - measure timing of execution specified function

As result, to open & login you need to write following:

go("http://${server}:${port}")
$('#user').sendKeys("${username}")
$('#pass').sendKeys("${password}")
$('#loginButton').click()

Measurement is also simple:

measure(function(){
    go("http://${server}:${port}/mypage.jsp?id=${some_id}");
})

2. Dev Mode

This mode allow to use alwasys open browser for scripting. 

3. Arguments 

Sometimes you need to run Chrome using specific flags. This patch allows to specify flags in Chrome Config

### DIFF
--- a/webdriver/src/com/googlecode/jmeter/plugins/webdriver/config/ChromeDriverConfig.java
+++ b/webdriver/src/com/googlecode/jmeter/plugins/webdriver/config/ChromeDriverConfig.java
@@ -5,6 +5,7 @@ import org.apache.jorphan.logging.LoggingManager;
 import org.apache.log.Logger;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.chrome.ChromeDriver;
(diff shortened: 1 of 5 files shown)


undera:
This PR is totally outdated. Let's close it and have new PR-per-feature if it is still relevant. I will review those PRs myself.
It makes sense to discuss suggested changes in mailing list prior to opening PR.

PR #41, opened 2014-03-11
Json extractor

phaas:
Single values extracted from a JSON array are wrapped in an array. This makes it difficult to use the extracted value in a subsequent request.

This change is to unwrap extracted values if they are arrays of size 1.

Note: This changes existing behavior.


undera:
Described jmeter-standard approach was implemented

PR #50, opened 2014-05-04
(feature) add mode to generate graphs for each sampler

Cka3o4Huk:
This micro-change is written to add ability to generate graph not only for whole jtl, but for each sampler. It's useful if scenario contains more than 5 samplers or you're interesting of particular sampler.


undera:
The project had major changes in the source code structure. Please bring new PR if needed.

Cka3o4Huk:
OK, I'll check if it's possible.

undera:
Actually now there is new Plugins Manager infrastructure that allows you to maintain your own plugin source code under your account and have it distributed through jp@gc.

PR #69, opened 2014-11-04
Selenium Grid Connection using a Remote WebDriver

jscruz:
Added a Remote WebDriver GUI so you can connect to a Selenium Grid.

In the GUI you need to insert your URL for the selenium grid (p.e http://seleniumgrid.com/wd/hub).
Then you can select a Capability (Firefox, Chrome, Internet Explorer)

In the future we could add another component for selecting the browser.


djangofan:
Thank you!  I can't wait to see this new feature in action.  For the longest time I have wondered why this driver wasn't the first one to be done since its far more useful than spawning a local browser instance.   I have plugin pack 1.2 and so I am assuming I need to wait in order to see this change?   Also, will this support a remote grid to SauceLabs ,which requires greater customization of the Capabilities object?

jscruz:
Hi @cplim, 

Thank you for your reply. 

I agree with you. I think there is no need to these features, so I'll remove the Proxy and User-Agent from the GUI. The unit test was done but I had a typo in the name, I'll fix it.

@djangofan, I've never used SauceLabs before but it seems pretty interesting. 

Thanks

Regards, 
Jorge

djangofan:
@cplim  SauceLabs is excellent.  You can get a 2 threaded free account.   To get Jmeter to run a sauce labs test, its just like any other RemoteWebDriver grid hub url, and so it is reasonable for you to add support for it.   All you need is to make sure that the Jmeter settings panel for the remote web driver have the following options (which also apply to a generic grid hub)  :
    0. Create a Capabilities object
    1.  field to add capability 'browserName' = "firefox"
    2.  field to add capability 'platform'  = "Linux"
    3.  field to add capability 'version'  = "33"
    4.  field to  add capability 'screen-resolution' = "1280x1024"
    5.  field to add capability 'name'  =  "Default test name " + ${__threadNum}
    6.  field to add capability 'build'  =  "firefox 33 on linux"
    7.  hard coded  capability 'selenium-version' = "2.44.0"  ( or whatever jmeter supports at the time)
    8.  provide a field for grid url
    9.  provide a field for saucelabs user
   10.  provide a field for saucelabs auth key

Then, the only extra thing you could OPTIONALLY do is to create a  RemoteWebDriver post-processor that would acquire the saucelabs sessionid and update the test name from each Sampler by implementing the simple SauceLabs rest-api  to  update the test name AFTER each sampler runs.  That way the jobs shown in saucelabs will no longer show the default name.

If you have any questions , email me, as I would love to help make it a reality:  djangofan -AT- gmail dot com

cplim:
Thanks for the pull request, Jorge. Much appreciated! :)

jscruz:
You're welcome :)

jubax:
Hi Jorge,

I just tried out your new plugin (had to compile it because there is not yet a release which contains it). It works pretty well. I am able to execute my scripts on a selenium grid (the grid is installed on a local server).

However, I have a feature request: I would like to be able to add arbitrary capabilities. Of course, there should be some text fields for standard capabilities (like suggested in the comment by djangofan). But I would also need a way to add non-standard capabilities (e.g. I'm currently using firefox_binary to use a special firefox binary install on all of our internal grid nodes).

Thanks for your plugin and let me know if I can help or if you need further information.

Regards,

Juergen

webdriver/src/com/googlecode/jmeter/plugins/webdriver/config/gui/RemoteDriverConfigGui.java
```diff
@@ -0,0 +1,144 @@
...
+
+	@Override
+	public void focusGained(FocusEvent e) {
+		// Nothing to do
+	}
+
+	@Override
+	public void focusLost(FocusEvent e) {
+		if(remoteSeleniumGridText.equals(e.getComponent()) && !isValidUrl(remoteSeleniumGridText.getText())){
+			messageDialog.show(this, "The selenium grid URL is malformed", "Error", JOptionPane.ERROR_MESSAGE);
```
undera: You should never ever do it like this. This kills user experience.

jscruz:
Hi @undera,

What would you recommend in that case?

Thank you.

undera:
I already committed the fix for that, there just a red-colored label that shows error message without blocking the UI.

PR #75, opened 2015-04-16
Multiple "ColorsDispatcher" implementations

echiu64:
Little bit of housekeeping:
 - updated maven plugins so they will work with Maven 3.1+
 - Made GraphRowSimple serializable to avoid copy/paste errors in JMeter
 - Changed ColorsDispatcher to be a factory and then different implementations will be returned depending on properties set. Moved original implementation into a class called CycleColors

New ColorsDispatchers
 - HueRotatePalette - takes a base colour and creates N variations through hue rotation, as well as M gradients of black->gray
 - CustomPalette - specify a fixed custom palette, when the automated CycleColors or HueRotate isn't good enough

Near full test coverage is included


undera:
This looks awesome!
Please add 1 missing thing: document your changes, explaining how to set properties and which  ColorsDispatchers available. And don't forget changelog record. It's all listed in our developer guide here: http://jmeter-plugins.org/wiki/DeveloperGuide/

echiu64:
On it's way... would be nice to preserve color selections each time we clear, not sure how viable that is though...

undera:
Thanks for the contribution!

PR #79, opened 2015-05-06
Apply format to csv file (AggregateReportGui)

mastojun:
When it generated AggregateReport csv of jtl files on console, header of csv had 'aggregate_report_error%', but error ratio was no percentage value.
I apply format to CSVSaveSErvice, using StatGraphVisualizer same as other saveGraphToCSV function.


undera:
Thanks for your initiative!

According to http://jmeter-plugins.org/wiki/DeveloperGuide/ you need to:
 1. Add changelog record
 2. Fix unit tests that for some reason failed after your change, see TravisCI build

mastojun:
I'm sorry for this fault.

I fixed the compile errors. I have did't know jmeter-plugin can't be compiled on jdk 1.7+.
I add changelog record.

I tried modify test code of edited block. but I don't know what is good way how to verify this change.

Thanks you for your comment.

PR #86, opened 2015-07-24
Fixed permission-denied-error on unix systems in LoadosophiaUploader

dimitrijber:
When a plugin is used to run on multiple environments, it is sometimes preferable that the storeDir is set to the system's default temp directory. I tried out not to set storeDir at all, but then i got a "premission denied" error because the code tries to set an trailing slash to an empty string resulting in "/" which is then passed to File.createTempFile("Loadosophia_", ".jtl", new File("/")); which results in an temp-file in the root directory. If run with unprivileged user, it will fail to create that file.

Why is it necessary to set storeDir? Why don't you use the system's temp folder in all cases?

### DIFF
--- a/site/dat/wiki/Changelog.wiki
+++ b/site/dat/wiki/Changelog.wiki
@@ -3,6 +3,7 @@
 == 1.3.1 <i><font color=gray size="1">planning</font></i>==
   * add connectTime field into Flexible File Writer
   * add [ConnectTimeOverTime Connect Times Over Time] listener to Extras set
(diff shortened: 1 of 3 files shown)


undera:
Changelog record is missing, according to http://jmeter-plugins.org/wiki/DeveloperGuide/

undera:
Setting store dir is important. For example, some systems have very limited temporary directory space, so they will need a way to change the place to store intermediary data.

dimitrijber:
Oh, sorry, I should have better first read the DevGuide :-) It's done now.

undera:
Thanks!

PR #90, opened 2015-11-10
Add command line Filter Results tool launch with CMDRunner tool

vdaburon:
Hi,
I add  FilterResults Tool launch with the "CMDRunner.jar --tool FilterResults"
Example : 
java  -jar %0\..\CMDRunner.jar --tool FilterResults --output-file filter.xml --input-file gestdoc_300sec.csv --include-label-regex true --include-labels "P[1-3].*" --start-offset 2 --end-offset 180 --success-filter true --save-as-xml true
This new tool need also modifications for the CMDRunner tool.


undera:
Please add some links to the wiki page from docs index and plugin set index. Otherwise people won't find it.

vdaburon:
I add the documentation but there is a conflict on the changelog wiki file

undera:
Thanks for contributing. I've added one more link from Start.wiki page

PR #92, opened 2015-11-24
Adding additional field options.

jhosmer:
Add fields grpThreads sampleCount errorCount responseHeaderSize responseSize URL to FFW.

### DIFF
--- a/standard/src/kg/apc/jmeter/reporters/FlexibleFileWriter.java
+++ b/standard/src/kg/apc/jmeter/reporters/FlexibleFileWriter.java
@@ -44,7 +44,9 @@ public class FlexibleFileWriter
             + "startTimeMillis endTimeMillis "
             + "responseTimeMicros latencyMicros "
             + "requestData responseData responseHeaders "
-            + "threadsCount requestHeaders connectTime";
+            + "threadsCount requestHeaders connectTime "
+            + "grpThreads SampleCount ErrorCount "
+            + "responseHeaderSize responseSize URL";
     private static final Logger log = LoggingManager.getLoggerForClass();
     private static final String OVERWRITE = "overwrite";
     private static final String FILENAME = "filename";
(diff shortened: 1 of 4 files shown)


standard/src/kg/apc/jmeter/reporters/FlexibleFileWriter.java
```diff
@@ -44,7 +44,9 @@
             + "startTimeMillis endTimeMillis "
             + "responseTimeMicros latencyMicros "
             + "requestData responseData responseHeaders "
-            + "threadsCount requestHeaders connectTime";
+            + "threadsCount requestHeaders connectTime "
+            + "grpThreads SampleCount ErrorCount "
```
undera: please use same style like other fields 
SampleCount=>sampleCount, ErrorCount=>errorCount

undera:
Thanks for contribution. According to http://jmeter-plugins.org/wiki/DeveloperGuide/ you need to change the doc page for FFW and also add a record to changelog about your improvement

jhosmer:
Wiki doc page updated (and cleaned up a little) for FFW.  Added note to Changelog.

undera:
The wiki page does not use GitHub markdown, it uses old Google Code markup syntax. That's why the file extensions are .wiki and not .md. Please use old formatting for the table. Otherwise it will become broken on the website.

standard/src/kg/apc/jmeter/reporters/FlexibleFileWriter.java
```diff
@@ -44,7 +44,9 @@
             + "startTimeMillis endTimeMillis "
             + "responseTimeMicros latencyMicros "
             + "requestData responseData responseHeaders "
-            + "threadsCount requestHeaders connectTime";
+            + "threadsCount requestHeaders connectTime "
```
undera: Here I see "threadsCount" while other places show "allThreads" isn't this a mistake?

jhosmer: The threadsCount field existed but there was no button for it in the GUI.  When adding the button label I used "allThreads" to be consistent with other listeners/plugins but I'll change it to match the internal field name of "threadsCount".

undera:
Thanks for contributing!

PR #96, opened 2016-01-27
Small but useful addition

d0k1:
JSyntaxtTextArea is used for script editing
Added some standart script context variables
Added simple way to pass cli args to phantomjs webdriver

### DIFF
--- a/webdriver/pom.xml
+++ b/webdriver/pom.xml
@@ -37,7 +37,7 @@
     </distributionManagement>
 
     <properties>
-        <selenium.version>2.47.0</selenium.version>
+        <selenium.version>2.49.1</selenium.version>
     </properties>
 
     <repositories>
@@ -150,6 +150,11 @@
             <groupId>de.sciss</groupId>
             <artifactId>jsyntaxpane</artifactId>
             <version>1.0.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fifesoft</groupId>
+            <artifactId>rsyntaxtextarea</artifactId>
+            <version>2.5.8</version>
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
@@ -289,7 +294,7 @@
                             <includeArtifactIds>
                                 selenium-support,selenium-api,selenium-remote-driver,selenium-ie-driver,selenium-firefox-driver,selenium-chrome-driver,selenium-htmlunit-driver,gson,
                                 guava,json,commons-exec,jsyntaxpane,htmlunit,htmlunit-core-js,nekohtml,cssparser,jetty-websocket,sac,httpclient,httpmime,httpcore,
-                                cglib-nodep,phantomjsdriver
+                                cglib-nodep,phantomjsdriver,rsyntaxtextarea
                             </includeArtifactIds>
                         </configuration>
                     </execution>
--- a/webdriver/src/com/blazemeter/jmeter/webdriver/phantomjs/PhantomJSDriverConfig.java
+++ b/webdriver/src/com/blazemeter/jmeter/webdriver/phantomjs/PhantomJSDriverConfig.java
@@ -7,6 +7,8 @@ import org.openqa.selenium.phantomjs.PhantomJSDriver;
 import org.openqa.selenium.phantomjs.PhantomJSDriverService;
 import org.openqa.selenium.remote.CapabilityType;
 import org.openqa.selenium.remote.DesiredCapabilities;
+import org.apache.jmeter.util.JMeterUtils;
+import org.apache.jmeter.threads.JMeterContextService;
 
 public class PhantomJSDriverConfig extends WebDriverConfig<PhantomJSDriver> implements ThreadListener {
     private static final long serialVersionUID = 100L;
@@ -34,6 +36,13 @@ public class PhantomJSDriverConfig extends WebDriverConfig<PhantomJSDriver> impl
         capabilities.setCapability(CapabilityType.PROXY, createProxy());
         capabilities.setCapability(PhantomJSDriverService.PHANTOMJS_EXECUTABLE_PATH_PROPERTY,
                 getPhantomJsExecutablePath());
+        String cliArgs = JMeterContextService.getContext().getVariables().get("phantomjs.cli.args");
+        
+        if(cliArgs!=null && cliArgs.trim().length()>0) {
+            String args[] = cliArgs.split(",");
+            capabilities.setCapability(PhantomJSDriverService.PHANTOMJS_CLI_ARGS, args);
+        }
+        
         return capabilities;
     }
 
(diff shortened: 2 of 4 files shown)


undera:
Did you see it broke the unit tests?

webdriver/pom.xml
```diff
@@ -37,7 +37,7 @@
     </distributionManagement>
 
     <properties>
-        <selenium.version>2.47.0</selenium.version>
+        <selenium.version>2.49.1</selenium.version>
```
undera: This change is not mentioned and should not be here. Bumping up Selenium dependency is whole different topic.

webdriver/src/com/blazemeter/jmeter/webdriver/phantomjs/PhantomJSDriverConfig.java
```diff
@@ -34,6 +36,13 @@ protected Capabilities createCapabilities() {
         capabilities.setCapability(CapabilityType.PROXY, createProxy());
         capabilities.setCapability(PhantomJSDriverService.PHANTOMJS_EXECUTABLE_PATH_PROPERTY,
                 getPhantomJsExecutablePath());
+        String cliArgs = JMeterContextService.getContext().getVariables().get("phantomjs.cli.args");
```
undera: Why variables from the context? This does not look right. CLI args meant to be taken at the beginning of the test, variables are overkill.

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
```diff
@@ -1,35 +1,30 @@
...
 import java.awt.event.ActionListener;
+import org.apache.jmeter.gui.util.JSyntaxTextArea;
+import org.apache.jmeter.gui.util.JTextScrollPane;
 
 public class WebDriverSamplerGui extends AbstractSamplerGui {
 
     private static final long serialVersionUID = 100L;
     private static final Logger LOGGER = LoggingManager.getLoggerForClass();
 
-    static {
```
undera: why have you removed this section?

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
```diff
@@ -139,15 +134,20 @@ public void actionPerformed(ActionEvent actionEvent) {
...
     }
 
     private JPanel createScriptPanel() {
-        script = new JEditorPane();
-        final JScrollPane scrollPane = new JScrollPane(script);
-        setScriptContentType("text/plain");
-        script.setFont(new Font(Font.MONOSPACED, Font.PLAIN, 14));
+    	script = new JSyntaxTextArea(20, 20, false);
+    	script.getFoldManager().setCodeFoldingEnabled(true);
+    	script.setFont(new Font("Hack", Font.PLAIN, 14));
```
undera: What is font "Hack" and why you think it will be present on all platforms?

webdriver/pom.xml
d0k1: May I ask you why or where I can read about it? I didn't want to modify pom, but selenium 2.47.0 "dislikes" modern versions of firefox(43.0.4) and phantomjs(2.0+). It just doesn't work at all, i.e. it  starts a browser up and stay stalled. Updating selenium solved that problem. that's why I did this

webdriver/src/com/blazemeter/jmeter/webdriver/phantomjs/PhantomJSDriverConfig.java
d0k1: I chose that way because - it is the easiest way for me to reach my goal - start jmeter and set up some "global" variables, among them I would pass cli args for phantomjs driver. Moreover I'd like to add some options for firefox, for example set a Xdisplay to use. (I want to start firefox on xvfb's display, not on primary X display in Ubuntu). And of course I can pass custom cli arg to phantomjs through passing that variable to JMeter's cli. For me it is very useful. And I would glad to know why  I am wrong.

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
d0k1: I changed "text area" of web driver's script. I did it because the default JComponent acting as a text area for script (actually I forgot it's name) did not deal with changing language of script. For example, if I chose groovy, java it throws exception and display a dialog any time I focus the editor. And It can't fold code, can't colorize the text (groovy or java I mean). So I used a component as JSR223 Sampler did - JSyntaxTextArea. I think it is greatest swing text editor component I even seen. So why to not use it? 
So I erased that static initializer.

May be you have any reasons to leave original text editor component?

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
d0k1: I'm sorry - my fault I'm going to add some checks of course. For example, if a platform has Hack font prefer it  - otherwise use any available monospace font. FYI  https://github.com/chrissimpkins/Hack.
Is it okay? or I should find another way to set preferred font for text area? May I use there JMeter's properties?

d0k1:
Thanks for your time for review PR!
And about broken tests. I got the problem so I will find a way to pass java.awt.headless=true

webdriver/pom.xml
undera: I know that there is issues with Selenium versions, but it's always the case. They release new browsers and new Selenium all the time. Updating the dependency is a different track.

webdriver/src/com/blazemeter/jmeter/webdriver/phantomjs/PhantomJSDriverConfig.java
d0k1: BTW, I saw new commit you merged. It adds a possibility to set cli args in UI. So my changes are meaningless. I'll drop them

undera: Please pull from master branch, there were another PR relating to this, which solves CLI args case in a better, more jmeter-standard way.

undera:
Please read http://jmeter-plugins.org/wiki/DeveloperGuide/

You have to provide doc updates and have unit tests passing. There are thousands of people using the plugins, we have to maintain quality level.

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
undera: I'm ok with changing the component, just not drop the headless case... Most likely Unit Tests broke because of this.

webdriver/src/com/googlecode/jmeter/plugins/webdriver/sampler/gui/WebDriverSamplerGui.java
undera: I'd just use Font.MONOSPACED constant. People are so subjective on fonts beauty, you know...  Let JVM settings to handle this. Using Jmeter property is an overkill IMAO.

undera:
The project had major changes in the source code structure. Please bring new PR if needed.

PR #97, opened 2016-02-09
Http Simple table server, no cache for Response add headers

vdaburon:
add header for response with no cache.
In the Changelog add links to documentation


undera:
Thanks!

PR #100, opened 2016-02-23
Simplified some boolean expressions and removed unrequired returns

ham1:


undera:
There are major changes in process, consider reading this: https://groups.google.com/forum/#!topic/jmeter-plugins/SYXTAWzwCRc

PR #103, opened 2016-03-30
Adding a poisson arrival thread group as part of Jmeter plugin

visahak:
Generate arrival rates based on a poisson process. This should allow for the creation of a more realistic load.  
The need to have poisson arrivals has been discussed in on the google jmeter-plugin forum.


undera:
Hi,

This is strong contribution. But people won't find it if there will be no documentation explaining how it works. Please add documentation article describing this plugin, having screenshot of it. Also add Changelog record. See http://jmeter-plugins.org/wiki/DeveloperGuide/ for more info.

visahak:
Thanks. will do.

undera:
How can user specify for how long to hold the reached level of the load after ramp-up?

visahak:
The lambda parameter defines how many requests per second you want on average. The arrival rates follow a poisson process from the beginning. 

The ramp-up time is the total time of the experiment, and the average load is the same from the beginning till the end. I guess `ramp-up` time may be misleading. I can change it. any suggestions?

undera:
The practical aspect is that users intent to have average rate increasing for some time and then constant for some time. That's their expectation of "ramp-up" and "hold for/constant". Without such ability they won't find plugin useful for their wide variety of use-cases.

From the code, I don't see such calculations happening to have increasing average and then constant time to hold it. I think this is very important to offer. Could you improve the plugin to support it?

visahak:
While I understand the expectation of users and the concept of "ramp-up" and "hold for/contstant",  that is not how arrivals based on Poisson Processes works. Once you specify a certain lambda in this thread group, then all your arrivals will follow that poisson process. The design of this thread group is similar to the jmeter thread group.
  The plugin can be potentially be modified to allow users to change the lambda over time, but that model is referred to as Non-homogeneous Poisson Processes. http://www.math.uah.edu/stat/poisson/Nonhomogeneous.html and should be considered a new feature.

undera:
Homogenous process is a subset of non-homogenous, right? Having this thread group non-homogenous would cover the original case and also would solve the expectations from the audience.

I don't insist too much. However, the time interval before next release will be at least couple of months, so you have plenty of it to improve this plugin.

visahak:
Yes it is. We can potentially have the same interface for both, so when you want the thread group to be homogeneous, you set lambda to be the same across all intervals.

undera:
Plugins Manager now allows you to publish and maintain this plugin independently, I suggest going forward this way. It will give you full control.

PR #106, opened 2016-04-01
Added new option to allow delayed connection or reconnection feature to

treilhes:
JMXMon.
Fixed a bug where all JMXMon samplers following a failing JMXMon sampler
were not created
Added test case, change changelog, contributors wiki page


undera:
Merged, thanks!

PR #111, opened 2016-06-21
Set the good version lib in the script shell

vdaburon:


vdaburon:
Correct trouble when launch 2 times the JMXMon with JMeter GUI : 
ERROR - kg.apc.jmeter.jmxmon.JMXMonSampler: The client has been closed

tools/table-server/src/main/resources/org/jmeterplugins/protocol/http/control/simple-table-server.cmd
```diff
@@ -27,9 +27,7 @@ setlocal
 
 cd /D %~dp0
 
-set CP=..\lib\ext\ApacheJMeter_core.jar;..\lib\jorphan.jar;..\lib\ext\jmeter-plugins-table-server-2.0.jar
+set CP=..\lib\ext\ApacheJMeter_core.jar;..\lib\jorphan.jar;..\lib\ext\jmeter-plugins-table-server-2.1.jar
```
undera: this will always cause the problem with forgetting to change the line. Right way is to enable resources filtering in Maven and put here the properties from Maven to be automatically resolved by mvn.

vdaburon:
I added the resources plugin and configure the filter to add the version for the generated library in shell scripts as you suggested me.

undera:
looks good. I will merge it and release next version.

nixos89:
Hi everyone, 
I have still this issue with
> ERROR k.a.j.j.JMXMonSampler: The client has been closed.

... happening in JMeter 5.2.1 when ran more than once in GUI mode (same as @vdaburon ). I've posted question on [Stackoverlow](https://stackoverflow.com/q/63582566/6805866) and [SQA StackExhange](https://sqa.stackexchange.com/q/45567/45856), but didn't get any answer. @undera is there any way to fix this? 
Thank you in advance.

PR #112, opened 2016-07-05
Added unit test that is expected to fail due to bug in JSON Path Extr…

devinbostIL:
I added a unit test that is expected to fail due to the bug that was mentioned here:
[https://groups.google.com/forum/#!topic/jmeter-plugins/jZWHKbSjlw4](url)

The test results are:
`Expected :{"color":"red", "price":19.95}`
`Actual   :{color=red, price=19.95}`


devinbostIL:
I just want to point out that the build server will likely fail because the new unit test is expected to fail due to the bug mentioned in my comment above.

PR #116, opened 2016-08-23
Added support for jbrowserdriver

Brimstedt:
https://github.com/MachinePublishers/jBrowserDriver


undera:
Thanks for the initiative.
But it is not enough to add new driver into remote capabilities. There should be a way to add it into GUI.
This requires some code similar to PhantomJS's config.

Also, I'll need a list of libraries that are required to be installed into /lib to enable its functioning. It will extend the list we have currently: https://github.com/undera/jmeter-plugins/blob/master/site/dat/repo/jpgc-plugins.json#L470

undera:
Also, these features must be documented. Please see developer guide for the project: https://jmeter-plugins.org/wiki/DeveloperGuide/

Brimstedt:
Thanks for the comments, 
Unfortunately it seems jbrowserdriver may not be fit for our requirements, so it's not certain Ill develop this further. 
Will close the pull request for now

PR #120, opened 2016-11-02
Add missing DbMon, JMXMon Plugin Type for JMeterPluginsCMD

vdaburon:
DbMon and JMXMon are missing in Plugin Type


undera:
Thanks!

PR #125, opened 2017-02-20
add directory-listing plugin

artem-fedorov:


plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private Iterator<File> iterator;
+    private List<File> list;
+
+    @Override
+    public void iterationStart(LoopIterationEvent loopIterationEvent) {
+        // TODO: independentListPerThread
+
+        try {
+            if (list == null) {
```
undera: you can put this `if` into one below

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
```diff
@@ -0,0 +1,168 @@
...
+                boolean isReRead = getReReadDirectoryOnTheEndOfList();
+                boolean isRewindOnTheEnd = getRewindOnTheEnd();
+
+                if (isReRead && isRewindOnTheEnd) {
+                    this.list = getDirectoryListing();
+                    this.iterator = list.iterator();
+                } else if (isRewindOnTheEnd) {
+                    this.iterator = list.iterator();
+                } else {
+                    return;
```
undera: What's this situation? Why we return without any action?

artem-fedorov: If list iterator at the end && (isRewindOnTheEnd == false) 

What we doing in this situation?

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
artem-fedorov: no, next `if` for situation when iterator at the end of list

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
undera: Stopping thread.

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.jmeter.threads.JMeterContextService;
+import org.apache.jmeter.threads.JMeterVariables;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.*;
+
+public class DirectoryListing extends ConfigTestElement implements LoopIterationListener {
```
undera: Name it DirectoryListingConfig, that would explain the class better

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryListing.java
```diff
@@ -0,0 +1,170 @@
...
+                    this.iterator = list.iterator();
+                } else if (isRewindOnTheEnd) {
+                    this.iterator = list.iterator();
+                } else {
+                    // if the end of list && !isRewindOnTheEnd
+                    return;
+                }
+            }
+
+        } catch (FileNotFoundException ex) {
```
undera: Do you really need to catch this exception?

artem-fedorov: yes, here we must catch this exception and logging it or throw RuntimeException (failed all test)

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryIterator.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.apache.jmeter.threads.JMeterContextService;
+
+import java.io.File;
+import java.util.Iterator;
+import java.util.List;
+
+public class DirectoryIterator {
+
+    private final DirectoryListingConfig config;
```
undera: bad decision. business logic class should not depend on its clients. It's client class that should depend on business logic class.
The solution is to make DirectoryIterator to hold all the flags as fields. Or have dedicated Configuration class. Or have bitmasks to not have too many methods and fields.

plugins/directory-listing/src/main/java/com/blazemeter/jmeter/DirectoryIterator.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public DirectoryIterator(DirectoryListingConfig config) {
+        this.config = config;
+        list = config.getDirectoryListing();
```
undera: No, it's this class that has to do directory listing.

site/dat/wiki/DirectoryListing.wiki
```diff
@@ -0,0 +1,16 @@
+= Directory Listing Config =
+
+<span class=''>[/?search=jpgc-directory-listing <i class='fa fa-download'></i> Download]</span>
+
+Directory listing config plugin offers opportunities that allows to set file path from source directory into jmeter-variable in iteration loop.
```
undera: opportunities vs capabilities

site/dat/wiki/DirectoryListing.wiki
```diff
@@ -0,0 +1,16 @@
...
+This plugin have some config options, that influence for the behavior:
+  * "Use full path" - is this flag does not selected plugin will be set sub-path from source path into destination variable;
+  * "Random order"  - shuffle listing in random order;
+  * "Recursive listing" - meaning, that plugin will be looking for all sub-folders in source directory;
+  * "Rewind on end of list" - if the flag is selected and an iteration loop has reached the end, the new loop will be started;
+  * "Re-read directory on the end of list" - re-read directory before starting new iteration loop;
+  * "Independent list per thread" - determines that listing will be shared for all threads or each thread will be have own local copy.
```
undera: "will have" or "will be having", never "will be have"

PR #126, opened 2017-02-21
add rotating-listener plugin

artem-fedorov:


plugins/rotating-listener/src/main/java/com/blazemeter/jmeter/RotatingResultCollector.java
```diff
@@ -0,0 +1,110 @@
...
+            LOGGER.warn("Incorrect value of max samples count: " + getMaxSamplesCount());
+            return Integer.MAX_VALUE;
+        }
+    }
+
+    public String getMaxSamplesCount() {
+        return getPropertyAsString(MAX_SAMPLES_COUNT, "");
+    }
+
+    public void setMaxSamplesCount(String maxFileSize) {
```
undera: Better rename parameter, as the meaning has changed

PR #129, opened 2017-02-28
Make installable files for plugins available in ZIPs

artem-fedorov:


downloads.py
```diff
@@ -18,38 +20,84 @@ def is_version_packed(fname):
...
 
-def pack_version(fname, ver_obj, pmgr_obj):
+def pack_version(fname, ver_obj, pmgr_obj, installer_cls):
     if not ver_obj['downloadUrl']:
         return
 
     with zipfile.ZipFile(fname, 'w', zipfile.ZIP_DEFLATED) as ziph:
-        # pack main file
-        download_into_zip(ziph, ver_obj['downloadUrl'], os.path.join("lib", "ext"))
+        if installer_cls is not None:
```
undera: Why this if is so high? Let's have single flow, only varying with call to installer as one of the steps...

downloads.py
```diff
@@ -18,37 +20,59 @@ def is_version_packed(fname):
...
-        # pack main file
-        download_into_zip(ziph, ver_obj['downloadUrl'], os.path.join("lib", "ext"))
+        tmp_dir = tempfile.mkdtemp()
+
+        # download main jar
+        jar_path = download_into_dir(tmp_dir, ver_obj['downloadUrl'], os.path.join("lib", "ext"))
 
-        # pack libs
+        if installer_cls is not None:
+            os.mkdir(os.path.join(tmp_dir, 'bin'))
```
undera: If this dir happens to be empty at the moment of zipping, remove it.

downloads.py
```diff
@@ -18,37 +20,59 @@ def is_version_packed(fname):
...
+            return_code = subprocess.call(["java", "-cp", jar_path, installer_cls])
```
undera: using "check_call" instead of "call" will eliminate the need in following `if`. It's ok to throw exception in case of hard-fail.

downloads.py
artem-fedorov: if directory is empty it does not archived in zip_dir function

PR #131, opened 2017-02-28
fix merge-results installation

artem-fedorov:


undera:
We need plugin descriptor to have version 2.1 for it, to update the users

undera:
BTW, Travis is failing...

PR #133, opened 2017-03-02
View version changes

artem-fedorov:


manager/src/main/java/org/jmeterplugins/repository/Plugin.java
```diff
@@ -340,6 +340,15 @@ public boolean isUpgradable() {
         return depends;
     }
 
+    public String getVersionChanges(String versionStr) {
+        Map<String, String> depends = new HashMap<>();
```
undera: What's this variable for?

artem-fedorov: it's need remove

site/dat/repo/self.json
```diff
@@ -47,6 +47,13 @@
         "libs": {
           "cmdrunner": "https://search.maven.org/remotecontent?filepath=kg/apc/cmdrunner/2.0/cmdrunner-2.0.jar"
         }
+      },
+      "0.12": {
+        "changes": "Added displaying information about changes in version; Added caching images; Added tooltips in versions, with version changes; Fix error changing the color to red.",
```
undera: Use `\n` to divide records. Use `- ` to form sort of bullets list.

PR #134, opened 2017-03-06
New cmd commands

artem-fedorov:


manager/src/main/java/org/jmeterplugins/repository/PluginManagerCMD.java
```diff
@@ -18,7 +16,8 @@
     protected int processParams(ListIterator listIterator) throws UnsupportedOperationException, IllegalArgumentException {
         LoggingManager.setPriority(Priority.INFO);
         if (!listIterator.hasNext()) {
-            throw new IllegalArgumentException("Command parameter is missing");
+            showHelp(System.out);
+            return -1;
```
undera: Why -1?

manager/src/main/java/org/jmeterplugins/repository/PluginManager.java
```diff
@@ -331,4 +325,91 @@ private String getAllPluginsStatusString() {
...
+    }
+
+    private String getAvailablePluginsString() {
+        final StringBuilder buf = new StringBuilder("Available Plugins\r\n");
+
+        for (Plugin plugin : getAvailablePlugins()) {
+            Set<String> versionSet = plugin.getVersions();
+            String[] versions = versionSet.toArray(new String[versionSet.size()]);
+
+            if (versions.length == 1) {
```
undera: Don't vary output for users, it will make any automation scripts crazy. Be lean.

PR #136, opened 2017-03-07
Configuration plugins source

artem-fedorov:


manager/src/main/java/org/jmeterplugins/repository/JARSourceHTTP.java
```diff
@@ -129,10 +130,37 @@ protected JSON getJSON(String path) throws IOException {
         }
     }
 
+    public JSON getRepositories(String path) throws IOException {
```
undera: Can we have it as `protected`?

artem-fedorov: yes, we can

manager/src/main/java/org/jmeterplugins/repository/JARSourceHTTP.java
```diff
@@ -129,10 +130,37 @@ protected JSON getJSON(String path) throws IOException {
         }
     }
 
+    protected JSON getRepositories(String path) throws IOException {
```
undera: Maybe return JSONArray if we promise to consumer that it will be JSONArray?

site/dat/wiki/RequirementsForURLContent.wiki
```diff
@@ -0,0 +1,65 @@
+= Requirements For URL Content =
```
undera: The file name should be different. Something like `PluginRepositoryDescriptorFormat.wiki`

manager/src/main/java/org/jmeterplugins/repository/JARSourceHTTP.java
artem-fedorov: yes, and then return JSONArray in getRepo()?

PR #142, opened 2017-03-15
New version of wssecurity plugin

tilln:


undera:
Hi,
There is "changes" field for version item, it is good practice to write that what's new to clarify for users the need to upgrade. Do yo want to add it?

undera:
Great, thanks!

tilln:
How about the (binary) zip packages? Would you still provide them? (Are they generated automatically?)

undera:
ZIP packages are generated by Jenkins that publishes website. It will recognize this situation fine.

PR #147, opened 2017-04-13
added Websocket Samplers plugin

ptrd:
As discussed with Andrey:
"..... So I encourage you to include your plugin into jmeter-plugins.org registry. To do so, you need to provide pull request ...." 
so here is my pull request.
Thanks for sharing this plugin with your community.

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -207,5 +207,22 @@
         }
       }
     }
+  },
+  {
+    "id": "websocket-samplers",
+    "name": "WebSocket Samplers",
+    "description": "Provides a number of samplers for load testing WebSocket applications.",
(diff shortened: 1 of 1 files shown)


undera:
Hi,
All looks good, but I'd modify plugin name into something more distinguishable from existing "WebSocket Protocol Support". Otherwise it will be a bit confusing. 
Maybe you can make your to be "WebSocket Samplers by Peter Doornbosch" and also modify existing to be named "WebSocket Sampler by Maciej Zaleski". That way, it will reflect existence of 2 separate plugins for users.

ptrd:
That's fine with me. Done.

undera:
Great! Let's go!

PR #148, opened 2017-04-15
changed websocket-samplers download URL from bitbucket to github, bec…

ptrd:
…ause the bitbucket download returns filename in quotes, which leads to failing install


ptrd:
Great. Works fine now. Thanks.

PR #156, opened 2017-05-28
Common base class for all monitoring plugins

tilln:
Several existing plugins such as JMXMon, DbMon, Perfmon use a common pattern for building the GUI as well as using a Collector class with a worker thread for collecting the monitoring samples (which use the Response Message field for the queried value).
This pattern should be generalized so that the above as well as future monitoring plugins can inherit the common behavior.
This pull request just adds those common classes (without refactoring existing plugins) for review by maintainers/contributors.

### DIFF
--- a/infra/common/src/main/java/kg/apc/jmeter/graphs/AbstractMonitoringVisualizer.java
+++ b/infra/common/src/main/java/kg/apc/jmeter/graphs/AbstractMonitoringVisualizer.java
@@ -1,0 +1,233 @@
+package kg.apc.jmeter.graphs;
+
+import java.awt.*;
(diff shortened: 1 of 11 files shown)


undera:
16 codacy warnings have to be addressed

undera:
Also, there are chunks of uncovered code, which look like important. Especially for infra part.

tilln:
I'm aware of the coverage/warning issues. However, these come from existing code I'm proposing this refactoring for, unfortunately. 
If I had more time, I'd re-write the unit tests as well, but at this stage all I've done is pull up the common code into a common parent, including the (poor) test code.

undera:
Thanks! Let's proceed

PR #160, opened 2017-06-08
SSHMon Samples Collector

tilln:
Collects performance metrics that are returned from a command remotely executed over an SSH connection.


tilln:
The two classes with the low unit test coverage have integration tests (which run against a local SSH server).

undera:
Hi,
Why make this part of main repo? Can it be your repo?
My current intent is to not grow the main plugins repo, it's too monstrous now. And for you it would be easier to maintain your plugin and faster to release updates.
So my suggestiou would be to move this to separate repo like jmeter-wssecurity and just bring the plugin descriptor.

tilln:
There's a few reasons:

1. Dependency on jmeter-plugins-cmn-jmeter (see #156).
As other plugins use that dependency, they should all be in sync in terms of using the same version to avoid conflicts. This would be easier to maintain as part of the main repo.

2. Help link on the title panel points to your wiki.
I'm using the base classes from #156 and they don't make it easy to customize the help link
(JMeterPluginsUtils.addHelpLinkToPanel() uses hard-coded URL parts via AbstractGraphPanelVisualizer.initGui() which is private).

3. Your repo seems the logical choice next to other, similar plugins (JMXMon, PerfMon, DbMon).
Also regarding consistent look and feel.

Let me know what you think.

undera:
I don't agree to these limitations. None of them should limit a developer from working on custom plugin. I'll explain reasons:
1. This dependency is not different than any other library dependency. Plugins Manager knows how to deal with it and there is version dependency possible to specify in plugins descriptor. As any good library, cmn changes very rarely and in compatible fashion. My intent is not to limit developers of plugins because of this library.
2. In case of separate repository, help link should point onto Project Page. And you are not obligated to provide help link at all. As a developer, you have full control to have that link and to point it to wherever you want. I want to benefit from developers not blocking on me reviewing their code change and feeling more responsible for their plugins.
3. For end users, it's mostly irrelevant where the source code is. They only care how useful and convenient the plugin is. Plugins Manager removes this complexity. If I'd have Plugins Manager in the past, all of mentioned plugin contributions would be separate repos. In my opinion, that would facilitate plugins development by not having a bottleneck of centralized code repo.

tilln:
Sure, fair enough, no worries.

> As a developer, you have full control to have that link and to point it to wherever you want.

I can't see how this can be done for sub-classes of ``AbstractGraphPanelVisualizer`` which already adds a link[ to the main repo wiki](https://github.com/undera/jmeter-plugins/blob/master/infra/common/src/main/java/kg/apc/jmeter/graphs/AbstractGraphPanelVisualizer.java#L119). 
Any suggestions?

undera:
Probably you have to just copy the piece of code, which is not too complex. Again, it's from the past, when I did not thought of Plugins Manager and today's reality of independent plugin vendors :)

tilln:
I was rather thinking of some proper base class design so that sub-classes can re-use functionality without having to copy pieces of code (which I'm not too fond of). 

For example, change that questionable pattern of adding the help link to the title panel without overriding the [inherited method](https://jmeter.apache.org/api/org/apache/jmeter/visualizers/gui/AbstractVisualizer.html#makeTitlePanel--), e.g.:

```
@@ -113,10 +113,20 @@ public abstract class AbstractGraphPanelVisualizer
     @Override
     public abstract String getStaticLabel();

+    @Override
+    protected Container makeTitlePanel() {
+        Container panel = super.makeTitlePanel();
+        String wiki = getWikiPage();
+        if (wiki != null) {
+            panel = JMeterPluginsUtils.addHelpLinkToPanel(panel, wiki);
+        }
+        return panel;
+    }
+
     private void initGui() {
         setBorder(makeBorder());
         setLayout(new BorderLayout());
-        add(JMeterPluginsUtils.addHelpLinkToPanel(makeTitlePanel(), getWikiPage()), BorderLayout.NORTH);
+        add(makeTitlePanel(), BorderLayout.NORTH);
         container = getGraphPanelContainer();
         container.add(createGraphPanel(), BorderLayout.CENTER);
         add(container, BorderLayout.CENTER);
```

External plugins could then simply not provide a `getWikiPage()` string.

tilln:
Alternatively, make the creation of the help link conditional in `JMeterPluginsUtils.addHelpLinkToPanel(Container panel, String helpPage)`

```
@@ -179,8 +179,12 @@ public abstract class JMeterPluginsUtils {
             return panel;
         }

-        if (helpPage.endsWith("Gui")) {
-            helpPage = helpPage.substring(0, helpPage.length() - 3);
+        // build link to help page unless it's already a URL:
+        if (!new org.apache.commons.validator.routines.UrlValidator().isValid(helpPage)) {
+            if (helpPage.endsWith("Gui")) {
+                helpPage = helpPage.substring(0, helpPage.length() - 3);
+            }
+            helpPage = WIKI_BASE + helpPage + "/?utm_source=jmeter&utm_medium=helplink&utm_campaign=" + helpPage;
         }

         JLabel icon = new JLabel();
@@ -190,7 +194,7 @@ public abstract class JMeterPluginsUtils {
         link.setForeground(Color.blue);
         link.setFont(link.getFont().deriveFont(Font.PLAIN));
         link.setCursor(new Cursor(Cursor.HAND_CURSOR));
-        link.addMouseListener(new URIOpener(WIKI_BASE + helpPage + "/?utm_source=jmeter&utm_medium=helplink&utm_campaign=" + helpPage));
+        link.addMouseListener(new URIOpener(helpPage));
         Border border = BorderFactory.createMatteBorder(0, 0, 1, 0, java.awt.Color.blue);
         link.setBorder(border);
```

tilln:
Closing this as obsolete. Hosting externally. PR #163 for plugins descriptor submitted.

PR #161, opened 2017-06-08
Migrate sense uploader to use backend listener

artem-fedorov:


plugins/sense-uploader/pom.xml
```diff
@@ -59,7 +59,6 @@
             <groupId>kg.apc</groupId>
             <artifactId>jmeter-plugins-emulators</artifactId>
             <version>0.2</version>
-            <scope>test</scope>
```
undera: Why this change? Emulators are not supposed to be used for anything except testing.

artem-fedorov: Replaced by `components` dependency

undera:
Did you test it well? And especially from non-gui environment, like Jenkins?

artem-fedorov:
Yes, I tested it on Jenkins, and then compare results at Sense and Jenkins Performance Plugin. They're equals.

PR #162, opened 2017-06-15
Make GUI help link URL creation conditional

tilln:
Most plugins have a help link on their title panel which is added via `JMeterPluginsUtils.addHelpLinkToPanel`.
However, this assumes that the help page is hosted on JMeter-Plugins.org and the URL is generated to point there, which does not work for external plugins.
In order for external plugins to still use the help link feature when subclassing common GUI classes (e.g. AbstractGraphPanelVisualizer etc) the help link should not be modified if it already is an (external) URL.

### DIFF
--- a/infra/common/src/main/java/kg/apc/jmeter/JMeterPluginsUtils.java
+++ b/infra/common/src/main/java/kg/apc/jmeter/JMeterPluginsUtils.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 public abstract class JMeterPluginsUtils {
     private static final Logger log = LoggingManager.getLoggerForClass();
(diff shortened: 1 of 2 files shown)


infra/common/src/main/java/kg/apc/jmeter/JMeterPluginsUtils.java
```diff
@@ -179,8 +179,12 @@ public static Component addHelpLinkToPanel(Container panel, String helpPage) {
             return panel;
         }
 
-        if (helpPage.endsWith("Gui")) {
-            helpPage = helpPage.substring(0, helpPage.length() - 3);
+        // build link to help page unless it's already a URL:
+        if (!helpPage.toLowerCase().startsWith("http")) {
```
undera: That will match any page starting with HTTP. You should use "http:// or https://" logic instead

undera:
There has to be unit test covering this change.

tilln:
@undera Do you really need those static checks fixed? This is tough when taking inherited or legacy code and needing to make minor changes (and may actually put off contributors). This PR actually increases the JMeterPluginsUtils coverage (by 0.95%) but it was low to begin with which I don't have time to fix unfortunately.

undera:
The reason I'm asking to cover changes is because I want to avoid simple things like NPEs to leak into codebase. My practice shows that simple coverage gives a lot of control for basic quality.
I hope contributors understand why I'm asking for it, it is only because we care of quality for our users.

This might be one more reason why you would like to have next plugin as your own repo - you don't have to have any unit tests there. But for base infra classes I want to ensure that next change won't break previous.

undera:
Thanks for contributing this change!

tilln:
Totally agree with the coverage requirement. It's just frustrating when checks fail because coverage of a class was already low before a small change that has unit tests.
Happy to contribute, and I wish I had more time to add more tests.

PR #165, opened 2017-06-26
Fix sshmon version number

tilln:


undera:
The CI is back live, so should be deployed pretty fast

tilln:
Awesome, thanks. It's already on the site. 
Cheers

PR #166, opened 2017-07-04
Blazemeter uploader plugin

artem-fedorov:


plugins/blazemeter-uploader/src/main/java/com/blazemeter/jmeter/BlazemeterUploaderGui.java
```diff
@@ -0,0 +1,204 @@
...
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+public class BlazemeterUploaderGui extends AbstractListenerGui implements HyperlinkListener {
+
+    public static final String WIKIPAGE = "BlazemeterUploader";
+    public static final String UPLOAD_TOKEN_PLACEHOLDER = "Replace this text with upload token received at a.blazemeter.com\nCan be used deprecated API keys or new improved keys.\nRemember that anyone who has this token can upload files to your account.\nPlease, treat your token as confidential data.\nSee plugin help for details.";
+
+    private JCheckBox anonymousTest;
```
undera: This gui element is not needed. Empty token means anonymous

plugins/blazemeter-uploader/src/main/java/com/blazemeter/jmeter/BlazemeterUploaderGui.java
```diff
@@ -0,0 +1,204 @@
...
+    private ActionListener actionListener;
+    private JCheckBox shareTest;
+    private JTextField testWorkspace;
```
undera: Workspace has to be implicit. User does not think of it. Whatever workspace he has with requested project/test - works. 99% users will have it single.

PR #168, opened 2017-07-11
Added plugin 'ElasticSearch Backend Listener'

delirius325:
_Not sure where or how I could add the plugin to the "Plugins manager", but I did find that ".json" file while scanning through the project._ 

The purpose of this plugin is to send sample results to an ElasticSearch engine to be able to monitor results through Kibana (or any other data visualizer that supports ElasticSearch). 

It also has support for build numbers if the tests are launched from Jenkins (only have to add a property "BuildNumber" with the %BUILDNUMBER%/$BUILDNUMBER variable).  This solution helps monitor response time over different builds.

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -390,5 +390,18 @@
         }
       }
     }
+  },
+  {
+    "id": "elasticsearch-backend-listener",
+    "name": "ElasticSearch backend listener",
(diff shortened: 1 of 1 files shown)


undera:
@delirius325 Currently, JAR file that is configured in "downloadUrl" option, does not have version suffix in filename. This makes Plugins Manager to always offer to upgrade it. Can you please upload a copy of same JAR into GitHub release, but with filename having Maven standard version suffix? It will help users avoid confusion.

delirius325:
Created PR to resolve issue.

https://github.com/undera/jmeter-plugins/pull/173

undera:
Thank you very much!

PR #171, opened 2017-08-07
Uniqueness support in STS

andreybpanfilov:
Hi,
we would like to use STS as a global queue for JMeter threads, unfortunately, STS does not allow to do that because it does not maintain uniqueness of data being inserted.


undera:
Ok, thanks

PR #175, opened 2017-08-28
Release pmgr 0.16

artem-fedorov:


undera:
There is one more place where 0.15 is mentioned. It is `out.xml` file. Please add change of it to this PR.

undera:
We'll wait for Taurus to make sure it works for them and release then. @greyfenrir will work on that from Taurus side.

PR #181, opened 2017-09-14
Updated version of plugin

delirius325:


undera:
Please, leave previous version for possible downgrades. Just add new version, like other plugins do

delirius325:
Changes done.

PR #200, opened 2017-12-02
Fix bug ReferenceError when using Groovy and use JSyntaxtTextArea ins…

pmouawad:
…tead of JEditor

This PR fixes the bug reported on google group.
Sampler has possibility to select an alternative to javascript, but when you select Groovy
an error occurs when selecting another element and then select WDS again, see attached screenshot.

Issue was due to DefaultSyntaxKit.initKit(). 
So I switched to the JSyntaxtTextArea component which is anyway more uniform.
Note that it would be nice to drop 2.13 support which would allow using JMeter core JSyntaxTextArea#getInstance() which will handle correctly the new Darcula LAF.

Regards
Philippe M. from UbikLoadPack Team

### DIFF
--- a/plugins/webdriver/src/main/java/com/googlecode/jmeter/plugins/webdriver/sampler/gui/JSyntaxTextArea.java
+++ b/plugins/webdriver/src/main/java/com/googlecode/jmeter/plugins/webdriver/sampler/gui/JSyntaxTextArea.java
@@ -1,0 +1,267 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


pmouawad:
screen shot 2017-11-22 at 09 16 30

undera:
Please review unit test failures

pmouawad:
Hi @undera,
Are you sure there are test failures on this one?

Or so you mean codacy which I fixed except for visibility for unit tests.

Thanks

undera:
There are NPEs in Travis log seen...

undera:
Thanks for contributing!

pmouawad:
Hi @undera,
Any release of plugins expected soon for webdriver and jmeter plugins manager ?

Thanks

undera:
It's easier with webdriver plugin, we can release it independently.
With Plugins Manager it is harder, since it is central part of infrastructure and we prefer to release it more rarely.
I'm on 2 week business trip anyway, so can't commit to anything right now.

pmouawad:
Hi @undera,
Regarding webdriver you should have a look at rultor, it’s very nice to release with it.

For jmeter-plugins-manager, What infrastructure is concerned ? The reported bug is really nasty as whenever network is slow, you end up having a freezed jmeter and need to kill it.

Regards

undera:
It's really more about me being busy, I'm in the middle of flights... I'll take a look once I have some time.

PR #201, opened 2017-12-02
Fix webdriver bundle

pmouawad:
Fix 404 on jna platform jar download


undera:
Sorry, did not see this PR before I fixed it myself...

PR #209, opened 2017-12-12
2.0.0 was after the 1.5.3 lib bracket

delirius325:
Fixed by moving comma and adding bracket


undera:
Looks like there is too much change, please remove changes that don't belong to you plugin

PR #230, opened 2018-02-18
WS Security Plugin Release 1.5

tilln:
New WSS Timestamp Pre-Processor
Add WSS UsernameToken timestamp precision option and no password option


undera:
If you have introduced new component, consider adding it to `componentClasses`, so auto-detection will work for it, too.

PR #243, opened 2018-04-10
More details when not enough threads

pmouawad:
Warning "No free threads available in current Thread Group" does not indicate which thread group is concerned.

Add this information.


pmouawad:
Issue reported by Codacy is false positive, it mixes java ThreadGroup with JMeter TG

PR #273, opened 2018-08-03
add bzm-rte plugin

rabelenda:


site/dat/repo/blazemeter.json
```diff
@@ -529,5 +529,29 @@
         }
       }
     }
+  },
+  {
+    "id": "bzm-rte",
+    "name": "RTE support",
```
undera: I'd make it "RTE Protocol Support" to be more specific

rabelenda:
I changed from `RTE support` to `RTE Protocols Support` to be more explicit. I decided to use Protocols instead of Protocol since, as far as I know, RTE is not a protocol,  but a set of protocols, in particular the plugin currently supports tn3270 and tn5250 which are IBM protocols based on telnet.

PR #289, opened 2018-09-28
Added Capability Type to ChromeDriverConfig

rvbernal:
Added Capability Type of UNEXPECTED_ALERT_BEHAVIOUR set to IGNORE to handle the Alerts, Prompts and Confirmations of the application under test.

### DIFF
--- a/plugins/webdriver/src/main/java/com/googlecode/jmeter/plugins/webdriver/config/ChromeDriverConfig.java
+++ b/plugins/webdriver/src/main/java/com/googlecode/jmeter/plugins/webdriver/config/ChromeDriverConfig.java
@@ -2,6 +2,7 @@ package com.googlecode.jmeter.plugins.webdriver.config;
 
 import org.apache.jorphan.logging.LoggingManager;
 import org.apache.log.Logger;
+import org.openqa.selenium.UnexpectedAlertBehaviour;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeDriverService;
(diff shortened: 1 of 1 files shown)


undera:
Are you sure this should be unconditionally set for everyone?

rvbernal:
Hi undera,

Just got the error when trying to handle alerts on the latest version of Chrome.

Tests are not executing on the latest version of Firefox and IE.

This might because of the Selenium version.

Thanks,
rvbernal

undera:
My point is that this change might not be desired for all the users.

rvbernal:
Understood, but I am a user of the plugin and in my test scenario, I need to deal with a lot of Alerts, Prompts and Confirmations to fulfill the form to be submitted to the server.

undera:
According to https://jmeter-plugins.org/stats/, there are 5700 other users of this plugin. I'd like to be sure that your change won't cause troubles for them. So maybe set this capability based on some JMeter property, or UI checkbox?

rvbernal:
Ok, there is no checkbox for Unhandled Alert Behaviour in UI. Understand the concern for the other 5699 users of the plugin, but the there are no failed Unit Tests, and I got hope that the change did not affect any existing functionalities. Anyways, thank you for entertaining.

jsa34:
Apologies for hijacking this thread, but tried messaging in the google groups a while ago and couldn't see a place to report "issues" - is there any news on a release date for the new plugin version?  At the moment some of the fixes that have been introduced since the last release a while ago fix some of the things that are stopping me running my tests on a remote grid at the moment, so any new would be much appreciated!

undera:
There is version 3.0 of plugin pretty fresh. And Google Groups is the right (and only) place of asking questions.

jsa34:
Apologies - you are absolutely right... The last release noted here on Github in the release section is months ago, but a newer release is available from the plugin hub.  Thanks :)

undera:
Looks stuck

PR #292, opened 2018-10-02
Update CaseFormatMode to get enum using map to improve performance

jmetertea:
Update CaseFormatMode to get enum using map to improve performance
Update wiki for better documentation


jmetertea:
Use explicit scoping instead of the default package private level

undera:
Why not deprecate this FN in favor of core? I mean this on: http://jmeter.apache.org/usermanual/functions.html#__changeCase

After core has obtained this fn, users are confused by two functions doing same thing

jmetertea:
@undera Core team didn't wanted to add **all** case options to __changeCase, 
So I created a plugin function __caseFormat with all case options

undera:
But why you brought the function into core then, if core team does not want to update it? Do you have any proof they don't want to update it? Maybe email from archive or PR on github?

I'm just trying to understand from user's perspective - which fn to use... And I see great confusion between core and plugin.

jmetertea:
The "proof" is inside enhancement https://bz.apache.org/bugzilla/show_bug.cgi?id=61759
For example Graham Russell comment on removing full options:

> I agree with removal.
> I can't see a use case for it, at least not one worth keeping the code in core. If someone really needs it they can write their own with the particular requirements they have.

undera:
Bummer... It's sad to see this mess happening. And I don't see any good solution that will decrease confusion for users...

pmouawad:
@jmetertea , @undera , if you see a real use case, then I would suggest proposing again to include it in core function.
I think the problem at that time is that no use case was described , so we ended up thinking it would introduce too much complexity for very low usage.

jmetertea:
@undera notice that LowerCase and UpperCase functions can be removed

undera:
It makes not sense to remove them, because many people rely on their existense. Also, they provide syntactic sugar.

PR #293, opened 2018-10-02
Update cmd plugins

artem-fedorov:
- Update cmn to 0.6
- Update cmdrunner to 2.2


undera:
Maybe change other cmdtools, too?

artem-fedorov:
Yes, let's change them too

PR #294, opened 2018-10-03
Redis exception when RANDOM_KEEP chosen on a list

adbatrockend:
* Allow user to specify whether the data is a Set or List, and invokes appropriate API call
* Remove reference to random, since this is not supported in Lists
* Update documentation


adbatrockend:
Basic unit tests added on retrieval from set and list, and check for data exhausted throwing appropriate thread-ending exception

PR #298, opened 2018-10-10
Upgrade to Selenium 3.14

bpapez:
Here is a new attempt to upgrade to Selenium 3. 
All unit tests are passing - except for PhantomJS, as I don't know how to setup on Windows. According to WebDriver PhantomJS is deprecated and should be replaced by Firefox/Chrome headless mode.

### DIFF
--- a/site/dat/repo/jpgc-plugins.json
+++ b/site/dat/repo/jpgc-plugins.json
@@ -728,6 +728,32 @@
           "platform": "https://search.maven.org/remotecontent?filepath=net/java/dev/jna/platform/3.5.2/platform-3.5.2.jar",
           "jmeter-plugins-cmn-jmeter": "https://search.maven.org/remotecontent?filepath=kg/apc/jmeter-plugins-cmn-jmeter/0.4/jmeter-plugins-cmn-jmeter-0.4.jar"
         }
+      },
+      "2.4": {
+        "changes": "Add vars, props and ctx to Sampler",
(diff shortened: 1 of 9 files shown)


bpapez:
I added a fix to make all unit tests pass.

undera:
Can you also prepare change for plugins.json, for releasing new version? I want to make sure I don't miss any library changes.

bpapez:
I added a jpgc-plugins.json to the pull request with the prepared library changes. 
I have removed gson, cglib-nodep, sac, jna and jna-platform , because I saw that they were dependencies in the previous selenium jars, but have been removed meanwhile.

I was however testing just with Chrome and Firefox. I tried with the HTMLUnit driver, but I do not know how exactly it works or should work and how I can set the browser there. So it would be good if someone who knows tests it to see if a jar is missing with that.

site/dat/repo/jpgc-plugins.json
```diff
@@ -728,6 +728,32 @@
           "platform": "https://search.maven.org/remotecontent?filepath=net/java/dev/jna/platform/3.5.2/platform-3.5.2.jar",
           "jmeter-plugins-cmn-jmeter": "https://search.maven.org/remotecontent?filepath=kg/apc/jmeter-plugins-cmn-jmeter/0.4/jmeter-plugins-cmn-jmeter-0.4.jar"
         }
+      },
+      "2.4": {
+        "changes": "Add vars, props and ctx to Sampler",
```
undera: Description here should tell about Selenium library upgrade

undera:
Ok, I'll take it from here.

PR #305, opened 2018-11-02
add redis data writer as post-processor for write scenario

eyotang:
Dear,

Sometimes, we have the scenario to collect the info from last batch requests, then run next requests, redis data writer is very important for it.


undera:
Hi,
I don't accept new plugins into main repository anymore. 
Instead, create your own GitHub project with your plugin, release it and provide PR with plugin descriptor, like this one: https://github.com/undera/jmeter-plugins/pull/297

eyotang:
Thanks，I will try it

PR #318, opened 2019-02-08
Fix Concurrency Thread Group very high CPU issue

pmouawad:
As reported in:

- https://groups.google.com/forum/#!topic/jmeter-plugins/biuDmDSprFk

Improvements on CPU for same test:
Before : Avg:14.5%, Max : 99.1%
After : Avg : 8.53%, Max : 37%

If we zoom on first Minute:
Before : Avg:40%, Max : 99.1%
After : Avg : 17%, Max : 37%

### DIFF
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractDynamicThreadGroupModel.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractDynamicThreadGroupModel.java
@@ -24,7 +24,7 @@ public abstract class AbstractDynamicThreadGroupModel extends AbstractThreadGrou
     public static final String STEPS = "Steps";
     public static final String ITERATIONS = "Iterations";
     public static final String HOLD = "Hold";
-    protected final Set<DynamicThread> threads = Collections.newSetFromMap(new ConcurrentHashMap<DynamicThread, Boolean>());
+    protected transient Set<DynamicThread> threads = Collections.newSetFromMap(new ConcurrentHashMap<DynamicThread, Boolean>());
     protected final ResultCollector logFile = new FlushingResultCollector();
     protected volatile boolean running = false;
 
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
@@ -44,11 +44,12 @@ public abstract class AbstractThreadStarter extends Thread {
             supplyActiveThreads();
         } catch (InterruptedException e) {
             log.debug("Interrupted", e);
+            Thread.currentThread().interrupt();
         }
         log.debug("Thread starter has done its job");
     }
 
-    abstract protected void supplyActiveThreads() throws InterruptedException;
+    protected abstract void supplyActiveThreads() throws InterruptedException;
 
     protected DynamicThread makeThread(long threadIndex) {
         boolean onErrorStopTest = owner.getOnErrorStopTest();
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
@@ -44,6 +44,7 @@ public abstract class AbstractThreadStarter extends Thread {
             supplyActiveThreads();
         } catch (InterruptedException e) {
             log.debug("Interrupted", e);
+            // "InterruptedException" should not be ignored (squid:S2142)
             Thread.currentThread().interrupt();
         }
         log.debug("Thread starter has done its job");
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
@@ -29,14 +29,14 @@ public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
     }
 
     /**
-     * If threads is empty we wait on condition for 1s max to be notified,
-     *  otherwise 
+     * If threads is empty we wait on condition for DEFAULT_TEMPORISATION millis to be notified,
+     * otherwise we wait for 1s
      */
     public void waitThreadStopped() { // FIXME: method named in confusing way
         long sleepTime = threads.isEmpty() ? DEFAULT_TEMPORISATION : MIN_CHECK_TIME;
         lock.lock();
         try {
-            condition.await(sleepTime, TimeUnit.MILLISECONDS); 
+             condition.await(sleepTime, TimeUnit.MILLISECONDS); // NOSONAR 
         } catch (InterruptedException e) {
             log.debug("Interrupted", e);
             Thread.currentThread().interrupt();
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
@@ -1,32 +1,51 @@
 package com.blazemeter.jmeter.threads.concurrency;
 
-import com.blazemeter.jmeter.threads.AbstractDynamicThreadGroup;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.apache.jmeter.engine.StandardJMeterEngine;
 import org.apache.jmeter.threads.JMeterThread;
 import org.apache.jmeter.threads.ListenerNotifier;
+import org.apache.jmeter.util.JMeterUtils;
 import org.apache.jorphan.collections.ListedHashTree;
 import org.apache.jorphan.logging.LoggingManager;
 import org.apache.log.Logger;
 
+import com.blazemeter.jmeter.threads.AbstractDynamicThreadGroup;
+
 public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
     private static final Logger log = LoggingManager.getLoggerForClass();
-    public static final int MIN_CHECK_TIME = 1000;
+    private static final long DEFAULT_TEMPORISATION = JMeterUtils.getPropDefault("dynamic_tg.temporisation", 10L);
 
+    public static final long MIN_CHECK_TIME = 1000L;
+    private final transient Lock lock = new ReentrantLock();
+    private final transient Condition condition = lock.newCondition(); 
+    
     @Override
     protected Thread getThreadStarter(int groupIndex, ListenerNotifier listenerNotifier, ListedHashTree testTree, StandardJMeterEngine engine) {
         return new ConcurrencyThreadStarter(groupIndex, listenerNotifier, testTree, engine, this);
     }
 
-    public synchronized void waitThreadStopped() { // FIXME: method named in confusing way
-        if (!threads.isEmpty()) {
-            try {
-                wait(MIN_CHECK_TIME);
-            } catch (InterruptedException e) {
-                log.debug("Interrupted", e);
-            }
+    /**
+     * If threads is empty we wait on condition for 1s max to be notified,
+     *  otherwise 
+     */
+    public void waitThreadStopped() { // FIXME: method named in confusing way
+        long sleepTime = threads.isEmpty() ? DEFAULT_TEMPORISATION : MIN_CHECK_TIME;
+        lock.lock();
+        try {
+            condition.await(sleepTime, TimeUnit.MILLISECONDS); 
+        } catch (InterruptedException e) {
+            log.debug("Interrupted", e);
+            Thread.currentThread().interrupt();
+        } finally {
+            lock.unlock();
         }
     }
 
+    @Override
     public int getNumThreads() {
         return (int) Math.round(getTargetLevelAsDouble());
     }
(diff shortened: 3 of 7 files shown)


plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
```diff
@@ -1,31 +1,49 @@
...
 import org.apache.jorphan.collections.ListedHashTree;
 import org.apache.jorphan.logging.LoggingManager;
 import org.apache.log.Logger;
 
+import com.blazemeter.jmeter.threads.AbstractThreadStarter;
+import com.blazemeter.jmeter.threads.DynamicThread;
+
 public class ConcurrencyThreadStarter extends AbstractThreadStarter {
     private static final Logger log = LoggingManager.getLoggerForClass();
+    private static final long DEFAULT_SHIFT_RAMPUP = JMeterUtils.getPropDefault("dynamic_tg.shift_rampup_start", 0L);
```
undera: This is one of features to be able to change it in runtime via property. It can't be static, but can be cached like `rampUp`.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
```diff
@@ -53,14 +70,32 @@ private long getPlannedConcurrency() {
...
             double stepLen = rampUp / (double) steps;
             return Math.round(stepSize * (Math.floor(timeOffset / stepLen) + 1));
         } else {
-            double slope = maxConcurr / rampUp;
+            double slope = (double) maxConcurr / rampUp;
             return Math.round(slope * timeOffset);
         }
     }
+
+    void checkNeedsPropertiesReloading(long now) {
```
undera: Maybe make it private and call in constructor to avoid code duplication?

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
```diff
@@ -53,14 +70,32 @@ private long getPlannedConcurrency() {
...
+            this.steps = owner.getStepsAsLong();
+            this.maxConcurr = Math.round(owner.getTargetLevelAsDouble());
+            this.lastCachedTime = System.currentTimeMillis();            
+        }
+    }
+
+    /**
+     * @return the lastCachedTime
+     */
+    long getLastCachedTime() {
```
undera: If this is for testing needs, it should have `protected` access and be exposed via `ConcurrencyThreadGroupExt` class

undera: or have `lastCachedTime` as protected and move this method to `ConcurrencyThreadGroupExt`

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroupTest.java
```diff
@@ -104,9 +107,24 @@ public void testStartNextLoop() throws Exception {
...
+    @Test
+    public void testCachingOfProperties() throws InterruptedException {
+        Object[] objects = createTestPlan();
+        ListedHashTree hashTree = (ListedHashTree) objects[0];
+        ConcurrencyThreadGroupExt ctg = (ConcurrencyThreadGroupExt) objects[1];
+        
+        ConcurrencyThreadStarter starter = new ConcurrencyThreadStarter(0, new ListenerNotifier(), 
+                hashTree, new StandardJMeterEngine(), ctg);
+        long lastCachedTime = starter.getLastCachedTime();
+        Thread.sleep(ConcurrencyThreadStarter.CACHING_VALIDITY_MS/2);
```
undera: please reformat the code so there are spaces around `/` and `*` in equasions

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
```diff
@@ -44,11 +44,12 @@ public void run() {
             supplyActiveThreads();
         } catch (InterruptedException e) {
             log.debug("Interrupted", e);
+            Thread.currentThread().interrupt();
```
undera: Why is this needed? I assume the `run()` method will just finish and corresponding thread will finish.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
```diff
@@ -1,32 +1,51 @@
...
 import org.apache.jorphan.collections.ListedHashTree;
 import org.apache.jorphan.logging.LoggingManager;
 import org.apache.log.Logger;
 
+import com.blazemeter.jmeter.threads.AbstractDynamicThreadGroup;
+
 public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
     private static final Logger log = LoggingManager.getLoggerForClass();
-    public static final int MIN_CHECK_TIME = 1000;
+    private static final long DEFAULT_TEMPORISATION = JMeterUtils.getPropDefault("dynamic_tg.temporisation", 10L);
```
undera: This property will need a note in documentation, so users and us will know it exists for the future.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
```diff
@@ -53,14 +70,32 @@ private long getPlannedConcurrency() {
...
             return Math.round(slope * timeOffset);
         }
     }
+
+    void checkNeedsPropertiesReloading(long now) {
+        if (CACHING_VALIDITY_MS > 0 && now - lastCachedTime > CACHING_VALIDITY_MS) {
+            this.rampUp = owner.getRampUpSeconds();
+            this.hold = owner.getHoldSeconds();
+            this.steps = owner.getStepsAsLong();
+            this.maxConcurr = Math.round(owner.getTargetLevelAsDouble());
```
undera: original algorighm has no rounding. I'd avoid introducing it unless there is good reason.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractDynamicThreadGroupModel.java
```diff
@@ -24,7 +24,7 @@
     public static final String STEPS = "Steps";
     public static final String ITERATIONS = "Iterations";
     public static final String HOLD = "Hold";
-    protected final Set<DynamicThread> threads = Collections.newSetFromMap(new ConcurrentHashMap<DynamicThread, Boolean>());
+    protected transient Set<DynamicThread> threads = Collections.newSetFromMap(new ConcurrentHashMap<DynamicThread, Boolean>());
```
undera: why remove `final`?

undera:
The general approach is good. Please address my minor comments.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractDynamicThreadGroupModel.java
pmouawad: transient cannot be final  IMU : http://blog.clempinch.com/transient-and-final-instance-variables-in-java/

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
pmouawad: getTargetLevelAsDouble() transforms an int into a double. I do it once to avoid redoing it later each time

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
pmouawad: I didn't find documentation for the other one "dynamic_tg.shift_rampup_start", in which file should it be documented ?

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
pmouawad: AFAIU :  "InterruptedException" should not be ignored (squid:S2142)

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroupTest.java
pmouawad: Will be in next commit

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
pmouawad: Will be in next commit

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
pmouawad: As you guessed later for JUNIT

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
pmouawad: I don't understand, ConcurrencyThreadGroupExt does not extend ConcurrencyThreadStarter.
Can you clarify ? 
Thanks

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
undera: Sorry, I don't understand this reason. It is not ignored, it is logged. But the situation does not require any action apart from logging, the thread will just finish its job naturally.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
undera: It's this file: https://github.com/undera/jmeter-plugins/blob/master/site/dat/wiki/ConcurrencyThreadGroup.wiki
I'll add there other relevant properties after merging this PR.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
pmouawad: This is what the squid:S2142 says: IMU we must interrupt to be correct:

> InterruptedExceptions should never be ignored in the code, and simply logging the exception counts in this case as "ignoring". The throwing of the InterruptedException clears the interrupted state of the Thread, so if the exception is not handled properly the fact that the thread was interrupted will be lost. Instead, InterruptedExceptions should either be rethrown - immediately or after cleaning up the method's state - or the thread should be re-interrupted by calling Thread.interrupt() even if this is supposed to be a single-threaded application. Any other course of action risks delaying thread shutdown and loses the information that the thread was interrupted - probably without finishing its task.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
undera: This can change the math afterwards. Can we keep the math, unless we have a bug there?
These changes affect so wide audience that I want to avoid changes as much as possible.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
undera: Probably I'm wrong and misunderstood the change. My bad

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
undera: Still I don't see how doing interrupt changes anything... The code is so small and simple, it has not finish the thread...
Ok, let's keep it for now.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarter.java
pmouawad: In fact I do the same as before but I cache. Do you want me to revert ?

undera: Yes, please

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
undera: This property also needs documenting

undera:
Thanks for all the fixes. Now I have much more confidence merging it.
Thanks for your patience!

BTW I will release the new version right away, but it won't make it to appear on BlazeMeter engines, because this plugin is pre-installed on engine images...

ping0uin:
Hi @undera ,
do you know when it will be available on the blazemeter engines? This is a high priority for us. We cannot use properly, and spread the solution in our company without this...
Thanks.

undera:
I have no idea when it will get into BlazeMeter. Try asking their support.

PR #319, opened 2019-02-24
Fix NPE due to transient fields

pmouawad:
This should fix the NPE reported.

Sorry for this issue, there was no JUnit test covering this, so it was not caught.

Regards
Philippe M.


undera:
Do you see a way to have JUnit test covering this? I see this change is also not covered by any test, we might get into trouble again...

pmouawad:
We could serialize/deserialize and check non nullity.
A better option would be to have integration test in remote mode.

But contribution are welcome on this.

Thanks

PR #329, opened 2019-04-01
Fixed HLS 1.3 release .json format

Baraujo25:
Fixed HLS 1.3 release .json format


undera:
I realized that your repo is private. It does not allow downloads for public users.
JMeter-Plugins.org only hosts Open Source plugins at the moment.

rabelenda:
Thank you for the notice, we have already fixed that.

PR #331, opened 2019-05-07
Added capability to set Y-Axis label on perfmon charts.

WhiteLightninggit:
Having possibility to label Y-Axis in performance tests tool making charts is a must. 

Otherwise each time report is sent, people are asking what are units on Y axis? Its According to my own experience with perf tests and this:

https://groups.google.com/forum/#!topic/jmeter-plugins/zjTwY38X_z8
https://groups.google.com/forum/#!searchin/jmeter-plugins/axis|sort:date/jmeter-plugins/C-mQuB_8PI8/Bdc6pNlyxNYJ
https://groups.google.com/forum/#!searchin/jmeter-plugins/axis|sort:date/jmeter-plugins/zjTwY38X_z8/s8pbcK-tmk8J

So I've added posibility to set Y-label using `--yAxisLabel <String>` option. If its not specified default String which is "Performance Metrics" like currently is used, so it shouldn't be breaking anything.

I made this change way to not disrupt other things, so if no --yAxisLabel is specified, it uses default value, which is "Performance Metrics" like now, otherwise this label is set.

### DIFF
--- a/tools/cmd/src/main/java/kg/apc/cmdtools/ReporterTool.java
+++ b/tools/cmd/src/main/java/kg/apc/cmdtools/ReporterTool.java
@@ -46,6 +46,7 @@ public class ReporterTool extends AbstractCMDTool {
                 + "--exclude-label-regex <true/false> " // filter samples label with regex
                 + "--start-offset <sec>" // filter samples on period time
                 + "--end-offset <sec>" // filter samples on period time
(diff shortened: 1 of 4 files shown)


undera:
Sorry, I somehow missed that useful PR

mrlogan:
They say better late then never, or something like that :)

PR #332, opened 2019-05-20
 Added plugin 'JMeter Listener pack'

lansio:
This plugin is Listener with ability to send result to ClickhouseDB. It could send each Sampler-transation or aggregated Sampler-transaction info (averaged times, count of Sampler transactions etc.)
In additional, this plugin contains Classes for InfluxDB and ElasticSearch with Samplers aggregation ability.
Project link https://gitlab.com/testload/jmeter-listener
Wiki link https://gitlab.com/testload/jmeter-listener/wikis/home


undera:
Welcome!

PR #346, opened 2019-07-25
Fix Maven download filenames

tilln:
Handle optional quotes around filename when downloading Maven libs.
See https://groups.google.com/forum/#!topic/jmeter-plugins/wJDHJkoc98o

### DIFF
--- a/downloads.py
+++ b/downloads.py
@@ -60,7 +60,8 @@ def download_into_dir(dirname, url, dest_subpath):
     resp = requests.get(url)
     assert resp.status_code == 200
     if 'content-disposition' in resp.headers:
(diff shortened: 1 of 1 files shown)


downloads.py
```diff
@@ -97,7 +98,7 @@ def get_pmgr(plugins_list):
     for repo_file in os.listdir(repo_dir):
         with open(os.path.join(repo_dir, repo_file)) as fhd:
             content = json.loads(fhd.read())
-            jsonschema.validate(content, schema)
+#            jsonschema.validate(content, schema)
```
undera: Why disabling schema validation?

tilln: By accident. (Didn't get my pip to work...)

PR #361, opened 2019-12-22
Mqmeter

JoseLuisSR:
Hi @undera 

I updated the mqmeter version fix screenshotUrl reference and remove lib that is not necessary to use mqmeter.

Thank you.

### DIFF
--- a/site/dat/repo/blazemeter.json
+++ b/site/dat/repo/blazemeter.json
@@ -737,7 +737,7 @@
     "id": "mqmeter",
     "name": "IBM MQ Support",
     "description": "A JMeter Plugin to put and get message on IBM MQ Queue, also publish message on Topic. It connect to MQ Server through server channel using ip address, port number, userID and password (if the channel has CHLAUTH rules).",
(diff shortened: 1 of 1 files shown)


site/dat/repo/blazemeter.json
```diff
@@ -746,11 +746,8 @@
       "co.signal.mqmeter.MQClientSampler"
     ],
     "versions": {
-      "2.1.0": {
-        "downloadUrl": "https://github.com/JoseLuisSR/mqmeter/releases/download/2.1.0/mqmeter-2.1.0.jar",
-        "libs": {
-          "com.ibm.mq": "https://search.maven.org/remotecontent?filepath=com/ibm/mq/com.ibm.mq.allclient/9.0.4.0/com.ibm.mq.allclient-9.0.4.0.jar"
-        }
+      "2.1.1": {
```
undera: When you release new version, please add it into "versions" instead of replacing. This way, users with old version will get nicer upgrade experience. Also, some users might want to downgrade plugin in case of incompatible changes, so having old version is very important.

undera:
No response from contributor

PR #375, opened 2020-04-01
Added Patch 3.0.1 for HLS plugin

RicardoPoleo:
With this PR we are adding the following fixes:
- Fixed the playlist reload to keep working after interrupting a test and starting a new one.
- Fixed rounding issue when calculating reload time to allow for finer reload times.
- We included a jenv file to fix the java version to be used in dev environments using jenv.

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -556,6 +556,10 @@
       "2.6.8": {
         "changes": "Fixed breaking issue with SSL conflicts",
         "downloadUrl": "https://github.com/delirius325/jmeter-elasticsearch-backend-listener/releases/download/2.6.8/jmeter.backendlistener.elasticsearch-2.6.8.jar"
+      },
+      "2.6.9": {
+        "changes": "Fixed breaking issue with SSL conflicts",
+        "downloadUrl": "https://github.com/delirius325/jmeter-elasticsearch-backend-listener/releases/download/2.6.9/jmeter.backendlistener.elasticsearch-2.6.9.jar"
       }
     }
   },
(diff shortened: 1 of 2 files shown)


site/dat/repo/various.json
```diff
@@ -636,6 +636,11 @@
       "com.ubikingenierie.jmeter.plugin.hls.gui.HLSSamplerGUI"
     ],
     "versions": {
+    "6.6.0": {
```
pmouawad, line 639: Can I ask you why the PR included changes on another plugin ?  Please provide a clean PR

site/dat/repo/various.json
```diff
@@ -651,6 +656,13 @@
     "vendor": "adrianmo",
     "markerClass": "io.github.adrianmo.jmeter.backendlistener.azure.AzureBackendClient",
     "versions": {
+      "0.2.1": {
```
pmouawad, line 659: Same here I guess

site/dat/repo/various.json
RicardoPoleo: Hello! I thought it was weird too but, the version that it's now in master contains those lines as well (I just copy-pasted it tbh).
As can be seen here: https://github.com/undera/jmeter-plugins/blob/master/site/dat/repo/various.json

site/dat/repo/various.json
RicardoPoleo: Same goes here. Or is my PR wrong in some way?

pmouawad: Yes I think so. 
Diff should only concern your parts of code not  parts that are not yours.
I think you sync your fork and make a new PR.

PR #408, opened 2020-07-20
Add Jmeter Prometheus plugin

GiovanniPaoloGibilisco:
Add Jmeter prometheus plugin as suggested here: https://github.com/johrstrom/jmeter-prometheus-plugin/issues/82

I've removed markerClass and componentClasses from the same as I'm not sure what they should refer to.

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -814,5 +814,21 @@
         ]
       }
     }
+  },
+    {
+    "id": "jmeter-prometheus",
+    "name": "Prometheus Listener Plugin",
+    "description": "A Jmeter plugin to expose sampled metrics to an endpoint to be scraped by Prometheus.",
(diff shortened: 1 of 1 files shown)


undera:
It's almost good. Please provide in `vendor` whatever name you like - it may be your company name, or your full name, or your nickname.
Also, you have to provide `markerClass` - it's a full Java class name of your main component, used to identify presence of plugin in Java classpath. In your plugin, `com.github.johrstrom.collector.SuccessRatioCollector` seem to be a valid option.

undera:
`componentClasses` list all of test elements in your plugin. This field is used for nice feature when use opens JMX file on a machine without your plugin, and plugin manager offers to install missing plugin. This field is optional, you can omit it.

GiovanniPaoloGibilisco:
I've added the vendor, marker and component classes

undera:
Looks good now. I'm merging it.
When you have new releases of your plugin, please consider creating another PR, adding new version to `versions` section.
Thanks for contributing!

PR #420, opened 2020-10-13
Citrix-Plugin Release 0.6.0

3dgiordano:
**Enhancements:**

* Improvements in the stability and performance.
* Increase in concurrent execution capacity.
* Improvements related to error information, information of status on Sampler result and detailed errors in log file
* Session windows is improved, better identification and position for better execution view
* New timeout parameters are incorporated (related to ica, connect, logoff and disconnect)
* The Logon Timeout parameter from the "Citrix Application Launcher" UI was removed
* The clause evaluation of "Citrix Application Launcher" now wait for the application start 
* Before disconnecting, a necessary logoff is performed to maintain a correct session state
* Hash handling precision is improved, parameters are also incorporated in case its precision needs to be changed
* Documentation related to debugging is improved
* The default interaction interval was changed to 1 second
* Texts in the plugin are corrected aligning them to the conventions currently used
* Better No-gui execution support

### DIFF
--- a/site/dat/repo/blazemeter.json
+++ b/site/dat/repo/blazemeter.json
@@ -938,8 +938,8 @@
       "com.blazemeter.jmeter.citrix.sampler.gui.CitrixSamplerGUI"
     ],
     "versions": {
-      "0.5.5": {
-        "downloadUrl": "https://github.com/Blazemeter/CitrixPlugin/releases/download/0.5.5/citrix-jmeter-0.5.5.jar",
+      "0.6.0": {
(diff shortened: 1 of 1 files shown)


undera:
Just replacing version number is not nice for user experience. Because users with older versions will get `N/A` version recognized for them.
Instead, add new version alongside with old one, just like all other plugins do.

3dgiordano:
Hi @undera, I had not realized my mistake. It has already been solved.
Thanks.

PR #426, opened 2020-11-24
Remove old LoggingManager and Logger

FSchumacher:
With JMeter 5.4 we want(ed) to drop the deprecated LoggingManager and `org.apache.log.Logger` classes. While testing the release, we found, that jmeter-plugins still rely on those classes and will break startup of JMeter (that is another bug :) but inside of JMeter).
This PR tries to remove all occurrences of LogginManager and org.apache.log.Logger and replaces those by loggers from SLF4J (which is used by JMeter currently).

**Breaking Changes**

Most of the classes used the loggers as private instance or class variables and should be OK to be changed. However there are a few classes, that expose the old Logger class by storing those in protected or public class variables. This PR changes that! The loggers are made private to hopefully help future developers and change the implementations to SLF4J.

### DIFF
--- a/graphs/graphs-additional/src/main/java/kg/apc/jmeter/vizualizers/BytesThroughputOverTimeGui.java
+++ b/graphs/graphs-additional/src/main/java/kg/apc/jmeter/vizualizers/BytesThroughputOverTimeGui.java
@@ -10,7 +10,7 @@ import org.apache.jmeter.samplers.SampleResult;
 
 public class BytesThroughputOverTimeGui
         extends AbstractOverTimeVisualizer {
-    //private static final Logger log = LoggingManager.getLoggerForClass();
+    //private static final Logger log = LoggerFactory.getLogger(BytesThroughputOverTimeGui.class);
 
     private static Method sentBytesMethod;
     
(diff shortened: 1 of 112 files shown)


undera:
I can tell that removal of old logging classes would have huge negative impact. Many plugins have no maintaners at all and will become unusable. Releasing _all_ plugins because of JMeter core breaking such a central dependency is big disruption for ecosystem. I strongly recommend reconsidering this change.

Please let me know your decision. In case of negative decision, I'll have to solve this on the plugins side somehow. In any event, I won't re-release all plugins right away.

pmouawad:
Hello @undera ,
We don't intend to break eco-system.
We intend to reintroduce the removed classes.
Still we would like to be able in near future to remove them, so if you can solve it on the plugins side it would be great.

Thanks
Regards

undera:
Sounds good @pmouawad. Basically, I will have to ship those classes via plugins-manager JAR, so the classes still will be present, at least for pmgr users. Not a big change, but important one to keep users from distraction.

undera:
I will merge this PR, but without releasing new versions.

FSchumacher:
Thanks for the fast response.

undera:
I have released pmgr 1.6 that ships old logging classes inside. This should help if core drops old logger classes.

PR #427, opened 2020-11-27
Adding OutlierDetection artifacts.

rbourga:
As this plugin uses functions new in Java 1.8, ensure you configure the Java level >= 1.8 in your Eclipse properties to build this plugin (jmeter-plugins uses 1.7 by default).
This new plugin acts as a low-pass filter to remove outliers on the right hand-side of the data results. This will give a distribution of results closer to a "Normal" distribution, which is more suitable for statistical analysis than a distribution with a right-tail.


undera:
Hi,
JMeter Plugins repository does not accept any new plugins. Instead, create your own repository (example: https://github.com/tilln/jmeter-formman), publish JAR for downloads and create PR to add your plugin into registry (example: https://github.com/undera/jmeter-plugins/pull/415/files)

rbourga:
Hi Andrey,
a quickie: now that the plugin is not part of the jmeter-plugin repo, how
do we need to initialise the "WIKIPAGE" string to point to the wiki in my
repo and not in the jmeter-plugins repo?

For instance, if I set
public static final String WIKIPAGE = "RightTailOutlierDetection";
it will be going to
http://jmeter-plugin.org/wiki/wikipage/RightTailOutlierDetection to load
the page and won't find anything!

Thanks for your help,
Robert

Le ven. 27 nov. 2020 à 18:51, Andrey Pokhilko <notifications@github.com> a
écrit :

undera:
Like this:
https://github.com/Blazemeter/jmeter-bzm-plugins/blob/3f8583f9ed29168a63ec0139321f2b816c7804bc/parallel/src/main/java/com/blazemeter/jmeter/controller/ParallelControllerGui.java#L14

Andrey Pokhilko

rbourga:
Hi Andrey,
Happy New 2021 year!

I have created my own repository (rbourga/jmeter-plugins-2 (github.com)
<https://github.com/rbourga/jmeter-plugins-2>), made a V.1.0 release and
added a JAR file in the release.
But GitHiub is not letting me create a PR as the structure of my repo is
different from the jmeter-plugins repo and it cannot see the differences.
How can I proceed from here for the Pull Request?

Thanks again for your help - I have a bit of a learning curve here as it is
the 1st time I am doing this!
 Robert

Le ven. 11 déc. 2020 à 15:02, Andrey Pokhilko <notifications@github.com> a
écrit :

undera:
The PR is usually done using fork of jmeter-plugins repo, creating new branch and changing just plugin descriptor file. So you have your plugin repo to develop the plugin AND a GitHub fork of jmeter-plugins repo, where you create branch per release to create PR with plugin descriptor.
You are almost there :+1: ! Make a GitHub fork and then create a branch on it and then commit to that branch the change in descriptor (`jpgc-various.json` I guess). Then create PR from it.

rbourga:
Creating new PR for various.json only.

rbourga:
Hi Andrey,

Thanks for merging the outlier plugin.
I am seeing that when I click on the "Help on this plugin" link, it is
taking me to "404 Page not found".
Apparently, this is because the url I put in the WikiPage string is being
automatically appended with "jmeter-plugins.org" as in:
404 Not Found (jmeter-plugins.org)
<http://jmeter-plugins.org/wiki/https:/github.com/rbourga/jmeter-plugins-2/blob/main/tools/outlierdetector/src/site/dat/wiki/RightTailOutlierDetection.wiki/>

This is the URL I have put in the source code:
// URL for project Wiki page
public static final String WIKIPAGE = "
https://github.com/rbourga/jmeter-plugins-2/blob/main/tools/outlierdetector/src/site/dat/wiki/RightTailOutlierDetection.wiki
";

Also, I have a mismatch in the version number (plugin is looking for 1.0.0
and jar is 1.0) which is causing the Plugin Manager to prompt the user to
always update the outlier plugin!

I will fix these 2 issues and build a new release after you tell me how to
fix the help url.

Thanks,

Robert.

Le ven. 11 déc. 2020 à 15:02, Andrey Pokhilko <notifications@github.com> a
écrit :

undera:
All of those links are controlled via plugin descriptor JSON and your plugin source code. Just use full URLs there beginning with `https://...`. Double-check that the link you are placing is able to open in browser.
Plugin manager is looking for version `1.0.0` because that is what you have specified in `versions`.

rbourga:
Hi Andrey,

I don't know if you can help me, but I am having some issues in getting the
help url in my plugin working.

For instance, in the source code, I am using an external url to
jmeter-plugins and have tried something as simple as:
this.add(JMeterPluginsUtils.addHelpLinkToPanel(makeTitlePanel(), "
HTTPS://somewhere-else.org/"), BorderLayout.NORTH);

The blue help link is displayed in the UI.
But when clicking on the link, it takes me to "
http://jmeter-plugins.org/wiki/HTTPS:/somewhere-else.org/".
Something has appended "http://jmeter-plugins.org/wiki/" to the url and I
cannot figure it out why/where.

Now, I compared with another plugin such as the "Merge" plugin (which is
using an internal help page to jmeter-plugins), and when clicking on the
link, it takes you to "
https://jmeter-plugins.org/wiki/MergeResults/?utm_source=jmeter&utm_medium=helplink&utm_campaign=MergeResults
".
This final url is expected as per the source code
in JMeterPluginsUtils.java. It will correctly build the final url for the
internal page to jmeter-plugings as per:

public static String buildHelpPageUrl(String helpPage) {
        try {
            if (helpPage.matches("[hH][tT][tT][pP][sS]?://.*")) {
                log.debug("Help page URL found, skipping building link to
"+WIKI_BASE);
                return helpPage;
            }
        } catch (PatternSyntaxException ex) {
            log.warn("Invalid regex", ex);
        }

        if (helpPage.endsWith("Gui")) {
            helpPage = helpPage.substring(0, helpPage.length() - 3);
        }
        return WIKI_BASE + helpPage +
"/?utm_source=jmeter&utm_medium=helplink&utm_campaign=" + helpPage;
    }

So would you know why we have the WIKI_BASE prefix ("
http://jmeter-plugins.org/wiki/") for external pages when the
buildHelpPageUrl function is not adding this prefix?

Thanks again,
Robert

Le ven. 8 janv. 2021 à 15:15, Andrey Pokhilko <notifications@github.com> a
écrit :

undera:
The prepend happens in `JMeterPluginsUtils.java`. But it only happens if URL does not match the regex.
Have you tried with lowercase `https://`? It should work.

rbourga:
Help link finally working!
This was because I was
inadvertently using jmeter-plugins-cmn-jmeter-0.4.jar instead
of jmeter-plugins-cmn-jmeter-0.6.jar.
Switching to 0.6 fixed the issue!
So I will submit a new version of the plugin with helplink and version
number corrections soon.
Thanks for your help.

Le jeu. 21 janv. 2021 à 15:19, Andrey Pokhilko <notifications@github.com> a
écrit :

PR #436, opened 2021-02-11
Add latest version of SSH Sampler

tilln:
This makes the current version of the [SSH Sampler](https://github.com/yciabaud/jmeter-ssh-sampler) plugin available via the Plugins Manager.

Note: I'm not the author of that plugin.


undera:
Thanks for helping to maintain this up-to-date!

PR #452, opened 2021-04-13
New ISO-8583 Sampler plugin version

tilln:
With correct JSON syntax ;)


undera:
I experience the same 404 locally, that is found by Travis CI. Please take a look.

PR #457, opened 2021-05-13
Release v1.2 Correlation Recorder

RicardoPoleo:
Now is possible to separate your Correlation Rules into groups improving your testing capabilities:

Having a different set of rules separated into groups ease the testing due to the possibility of enabling and disabling them in bulk (applying its effect to the rules each contains)
Each group is self-managed, allowing to add, delete, disable and edit a set of rules without altering the rest of the correlations.
Other changes:

Now each Correlation Extractor/Replacement hide by default its advanced fields, displaying only the most used ones for a faster setup.
Improved the addition of custom extensions to your environment, by selecting "More" on each combo, a dialogue will allow you to interactively choose which extension want to use or remove.


undera:
Brackets misplaced

RicardoPoleo:
Hello @undera, 

I need to update some libraries since the ones that I added to that release are breaking JMeter. Can I do it without changing the version? Like, in this PR or another?

Hope to hear from you soon,

Thanks

undera:
you can in another PR

RicardoPoleo:
Okas. Thanks!

PR #478, opened 2021-08-23
Added DI-KafkaMeter plugin

rollno748:
Adding D-KafkaMeter plugin to extend Jmeter to performance evaluate Apache kafka

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -1226,5 +1226,25 @@
         ]
       }
     }
-  }
+  },
+  {
+    "componentClasses": [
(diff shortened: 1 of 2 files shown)


undera:
Looks like the repository of di-kafkameter is private. Thus, the JAR file is unavailable for download. Can you make the repository public?

rollno748:
Apologies, I have made it to public now.
Should be accessible now.



Thanks,
Mohamed Ibrahim

rollno748:
Should I create a new pull request ?



Thanks,
Mohamed Ibrahim

PR #481, opened 2021-08-27
Fixed dependency library for pubsub-1.1 plugin

rollno748:
Added Library which was missing during the plugin installation for GCP-Pubsub-1.1
Fixed image for pubsub


undera:
Please remove all the unrelated changes from your PR. Please stick the change to the plugin you own.
A little advice is to always create a separate branch per-PR, this way it is easier to fix everything, without recreating PRs.

PR #482, opened 2021-08-30
Fixed dependency library for PubSub plugin -1.1

rollno748:
Added Library which was missing during the plugin installation for GCP-Pubsub-1.1 Fixed image for pubsub


rollno748:
Missed the escaping seq

PR #484, opened 2021-08-30
Added dependency libs for Pubsub-1.1

rollno748:
Added missing library for pubsub-1.1


undera:
If you would look here: https://github.com/undera/jmeter-plugins/pull/484/files
you'd see that the diff still contains lines with whitespace changes, belonging to plugins other than pubsub. Please leave that whitespace as-is. Only change the part that belongs to your plugin.

rollno748:
Apologies, Thought that json structure will not cause any issue. 
Will create a new PR with my plugin change alone 

Thanks

PR #493, opened 2021-10-15
New version of ResultsComparator plugin

rbourga:
Addition of a Pass/Fail criteria for comparing a test run against a baseline.
User can now set a threshold for Cohen's d and all samplers having their d above this threshold will be marked as "failed", meaning that there is a too big degradation in the response times.


rbourga:
Thanks Andrey.
Fixed the issues.

rbourga:
Hi Andrey,
This time should be the good one - never believe copy-paste! (I lost an " when pasting...).

PR #517, opened 2022-03-03
HLS Sampler: jackson-jaxb-annotations may require jaxb-api in some cases

jloisel:
When another JMeter plugin auto-registers Jackson modules, it fails with exception:

> Uncaught Exception java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlElement in thread Thread[StandardJMeterEngine,5,main]. See log file for details.


jloisel:
Thanks! :+1:

PR #531, opened 2022-05-18
STS 3.1 with new features

vdaburon:
New features : 
Use groovy script and not beanshell script
Daemon mode for nohup linux command (STS process background linux)
Load files when STS startup
Set charset to read and  write file
Set charset to html response
SAVE : could force the ADD_TIMESTAMP value
CONFIG : display STS configuration
Change Maven configuration for JMeter Maven Plugin


undera:
CI fails for it.

BTW Vincent,
Maybe it would be more convenient for you to take this plugin into your own repo? That would give the credit to the real author of the plugin.

vdaburon:
> _Maybe it would be more convenient for you to take this plugin into your own repo_
Did i need to change the package ?
org.jmeterplugins.protocol.http.control to somethink likes 
io.github.vdaburon.jmeterplugins.protocol.http.control

undera:
You don't have to change the package. And you _should_ keep it for compatibility. That domain is retired anyway :)

vdaburon:
Did i need to declare the new version STS 3.1 in the site\dat\repo\jpgc-tools.json file ? or you do it ?

undera:
The issue is that I'd now have to re-publish plugins to Maven and change the jpgc-tools descriptor. Which is a bit disruptive for me, sorry.
That's why I suggest actual plugins owner to fully control the process of release, with just my confirmation of descriptor PRs...

PR #534, opened 2022-05-20
Publish JUnit Reporter 1.0

tilln:


undera:
I have a question: why does this plugin use own KPI calculator inside, instead of relying on [Backend Listener](https://jmeter.apache.org/usermanual/component_reference.html#Backend_Listener)?

IMO that would be more standard way for JMeter

tilln:
> I have a question: why does this plugin use own KPI calculator inside, instead of relying on [Backend Listener](https://jmeter.apache.org/usermanual/component_reference.html#Backend_Listener)?

It seemed a more flexible approach.
The plugin uses ["storeless"](https://commons.apache.org/proper/commons-math/javadocs/api-3.3/org/apache/commons/math3/stat/descriptive/StorelessUnivariateStatistic.html) implementations for one-pass calculations rather than keeping values in memory. 

I may have missed something though, so happy to refactor if someone points out how to implement a more "standard way" :-)

undera:
A more standard way would be exactly a `BackendListenerClient` implementation. I am just suggesting, though, not requiring it.
If you still want to keep the current implementation as-is -- so be it. Just confirm it and I'll merge the PR.

tilln:
Ok I see what you mean, thanks. I'll check it out for the next release.
In the meantime, let's just publish as it is for now.

PR #547, opened 2022-10-11
__strReplaceRegex instead of __strReplace

jloisel:
Key of the StrReplaceRegex function seems incorrect, it should be "__strReplaceRegex", not "__strReplace".


undera:
The function name would have to remain the same for the historical reasons. It was like that in the beginning and cannot change in a breaking way now.

jloisel:
The function is not working at all due to to wrong name (which is already used by another one). You're not breaking anything since it's already broken.

jloisel:
`__strReplace` and `__strReplaceRegex` have both the same key `__strReplace`.

undera:
Wow, that changes everything. Thanks for reporting this!

undera:
New version has been released, thanks for being persistent

PR #551, opened 2022-11-02
update atakama variabilization plugin

RASAMIZANANY:


site/dat/repo/various.json
```diff
@@ -1534,12 +1534,12 @@
...
-  {
+   {
     "id": "jmeter-atakama-variabilization-plugin",
     "name": "Atakama Variabilization Plugin",
-    "description": "Variabilization plugin for jmeter",
-    "screenshotUrl": "",
-    "helpUrl": "https://github.com/atakama/jmeter_plugin_variabilization",
+    "description": "Variabilization plugin for jmeter: https://github.com/atakama/jmeter_plugin_variabilization",
+    "screenshotUrl": "https://user-images.githubusercontent.com/2484866/182579087-59225ac5-8abb-4a87-835a-2f202b1374fd.png",
+    "helpUrl": "https://ph-eval.atakama-technologies.com/downloadWebsite/index.php/product/show?id=1",
```
undera, line 1542: This is not help page, it's a promotional landing page. Let's not confuse users of plugins.

site/dat/repo/various.json
```diff
@@ -1551,47 +1551,14 @@
...
-        "depends": [
-          "jmeter-core",
-          "jmeter-components"
-        ]
-      }
-	  ,
-	  "1.0.2": {
-		"changes": "Fix bug when there are parameters with empty values",
-        "downloadUrl": "https://github.com/atakama/jmeter_plugin_variabilization/releases/download/1.0.2/jmeter-atakama-parameters-1.0.2.jar",
+      "1.1.0": {
```
undera, line 1554: I recommend you keeping old version info for your plugin, and add new versions. Look how other plugins do it.

This is important to allow users to downgrade to older version of plugin in case something is broken.

undera (changes requested):
Please address my comments

site/dat/repo/various.json
RASAMIZANANY: Up this new  version, it will be the case. We want the users  to upgrade to this version and start using  the plugin from it

site/dat/repo/various.json
RASAMIZANANY: There is a video link  with how it works also. I can put it within the description?
What do you suggest?

site/dat/repo/various.json
RASAMIZANANY: We have changed or license system so that's why we have removed the older versions

site/dat/repo/various.json
RASAMIZANANY: We can also add a documentation link within this promotional landing page

undera: I think it would be right to put your landing page into description, where users supposed to get general information about the plugin. The help link should serve exactly the technical help page.

site/dat/repo/various.json
undera: IMO it's a very unfriendly strategy to cause the inconvenience to users. Users should be motivated by value you provide, not the artificial limitations you set. 
That said, I won't insist here, up to you.

PR #557, opened 2022-12-16
jpgc-webdriver: addition of webdriver for Edge and upgrade to Selenium 4.7.2

rbourga:
Update to jpgc-webdriver plugin.


undera:
I have published the JAR into maven, but actually CI complains that `opentelemetry-sdk-common` is not downloadable. I wonder why we need that library at all...

rbourga:
Hi Andrey,
Thanks for the merge.
However, plugin manager is still showing version 4.6.0 for the Selenium
plugin instead of 4.7.2.
How long must we wait before 4.7.2 appears online?
Thanks.

Le lun. 19 déc. 2022 à 16:02, Andrey Pokhilko ***@***.***> a
écrit :

undera:
It's already there: https://jmeter-plugins.org/?search=jpgc-webdriver

undera:
Don't forget there is local repo caches in pmgr

rbourga:
All good now, I can see 4.7.2 from Plugin Manager.
I will post a message on Linkedin JMeter group to inform users of the new
driver for Edge.
Thank you Andrey.

Le lun. 19 déc. 2022 à 16:47, Andrey Pokhilko ***@***.***> a
écrit :

PR #558, opened 2022-12-16
Partial fix for Throughput Shaping Timer

pbaykalov:
This PR partially fixes erratic behaviour of Throughput Shaping Timer with low requested intensity.

This is the problem: 
TST_MCVE
[TST_MCVE.zip](https://github.com/undera/jmeter-plugins/files/10245065/TST_MCVE.zip)

Actual intensity might strongly depend on what exactly the requested intensity is and how many threads there are.

The timer is still not guaranteed to work correctly in specific but realistic situaitons.

### DIFF
--- a/plugins/tst/src/main/java/kg/apc/jmeter/timers/VariableThroughputTimer.java
+++ b/plugins/tst/src/main/java/kg/apc/jmeter/timers/VariableThroughputTimer.java
@@ -167,8 +167,14 @@ public class VariableThroughputTimer
         if (log.isDebugEnabled()) {
             log.debug("Calculating {} {} {}", millisSinceLastSecond, cntSent * msecPerReq, cntSent);
         }
+        
+        
         if (millisSinceLastSecond < (cntSent * msecPerReq)) {
-            // TODO : Explain this for other maintainers
+            //this code allows for very big number of threads to be fired each second (at most 1 divided by how much time "delay" runs because it's synchronized)
+            //each second has number of requests to fire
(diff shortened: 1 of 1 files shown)


pbaykalov:
What's the main reason for `return 0;` in `delay()`?

undera:
> What's the main reason for `return 0;` in `delay()`?

It's the default branch, in case we're not delivering enough hits. The meaning of `0` here is "do the next sample as soon as possible, don't add any delay".

pbaykalov:
> > What's the main reason for `return 0;` in `delay()`?
> 
> It's the default branch, in case we're not delivering enough hits. The meaning of `0` here is "do the next sample as soon as possible, don't add any delay".

It's not "default" it's the only return statement in `delay`.

undera:
> > > What's the main reason for `return 0;` in `delay()`?
> > 
> > 
> > It's the default branch, in case we're not delivering enough hits. The meaning of `0` here is "do the next sample as soon as possible, don't add any delay".
> 
> It's not "default" it's the only return statement in `delay`.

Ok, maybe we're talking about different places in code. Can you point to exact file and line that you're talking about?

undera:
> > > What's the main reason for `return 0;` in `delay()`?
> > 
> > 
> > It's the default branch, in case we're not delivering enough hits. The meaning of `0` here is "do the next sample as soon as possible, don't add any delay".
> 
> It's not "default" it's the only return statement in `delay`.

Ok, I see what you mean. So, the meaning of it is we tell JMeter that there should be no delay at all on its side, since our code has already did introduce the required delay with `wait()`

PR #568, opened 2023-02-15
Correlation Recorder v1.2.3

3dgiordano:
- Improved compatibility with new versions of JMeter
- Compatibility problems with MS Windows are solved

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,37 @@
+name: Build Automation
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "*" ]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Setup PHP Action
+      uses: shivammathur/setup-php@2.24.0
+      with:
+        php-version: '5.6'
+        tools: composer
+
+    - name: Prepare upload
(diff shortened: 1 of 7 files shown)


undera:
I have created an issue on their support forum: https://travis-ci.community/t/unable-to-set-up-php-5-6-with-phpenv/13592

3dgiordano:
Looking at the logs, I see that the new images do not have pre-installed the versions that they indicate to have

Preinstalled PHP versions on the actual docker image of travis ci (Xenial)
  7.4
* 7.4.30 (set by /home/travis/.phpenv/version)

Some days ago (Xenial)
  5.6
  5.6.40
  7.1
  7.1.27
  7.2
* 7.2.15 (set by /home/travis/.phpenv/version)

In both cases it uses the Xenial distribution, but it seems that the latest version does not have those pre-installed versions.

undera:
Plugins website needs PHP 5.6 to work properly. The fact that TravisCI have broken the compatibility with older versions, while their docs still promise those versions -- overall very sad.

undera:
Or we need to find a way to configure the build to install PHP 5.6 from some PPA

pmouawad:
Hello @undera , 
Would witching to an up to date PHP version be too much work  for the website ?
5.6 is in EOL since many years so it can explain why Travis CI stopped supporting it.

Thanks

undera:
The hosting we use has quite old platform with PHP 5.6. It would be a lot of trouble to reinstall everything including OS. I'd rather change Travis for something else...

PR #571, opened 2023-02-28
Release UbikLoadPack Video Streaming plugin 8.0.7

pmouawad:
Hello @undera,
This PR is to publish a new version of UbikLoadPack Video Streaming plugin and drop previous version.

Thank you in advance for your help and always for your reactivity !

Regards
UbikLoadPack Team


pmouawad:
Hello @undera , I see tests are failing but it does not seem related to this PR.

Regards

undera:
I'm really puzzled by why it fails now. I could not find any quick answer to that. It looks like some issue with locale or so, 'cause there are exceptions on getting resource strings.

undera:
I tried to narrow down the problem in #573 , but no luck so far. It has NPEs in a place where it should not. I don't have spare time right now to fix it, maybe on weekend.

undera:
The whole process of releasing the change is done by CI. Even after merge, broken CI will not publish the changes.

undera:
Please pull from master and let's see if it is fixed

PR #581, opened 2023-04-03
Update of webdriver to version 4.8.3.1

rbourga:
Webdriver 4.8.3.1 includes:
- new tab to configure and/or customise specific browser extensions;
- upgrade to Selenium 4.8.3.


undera:
Is jackson-databind library to be included into JSON descriptor?

rbourga:
I have been able to test the new tab successfully without any additional libraries.
So apparently the jackson-databind library is not required at runtime; only at compile time.

PR #592, opened 2023-05-17
Update AutoStop.java

ricardasomni:
Add environment variable which indicates that threshold limit was hit and test was stopped. This is required later to check in Tear down thread group and act accordingly (e.g. in my case I want to send Slack message about failed test).

### DIFF
--- a/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
+++ b/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
@@ -268,6 +268,7 @@ public class AutoStop
 
     private void stopTest() {
         stopTries++;
+        System.setProperty("auto_stopped", "true");
 
         if (JMeter.isNonGUI()) {
             log.info("Stopping JMeter via UDP call");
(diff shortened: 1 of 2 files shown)


undera:
For this to complete, you'd need to also document this feature, so other people would be able to discover it, too. It lives in this file: https://github.com/undera/jmeter-plugins/blob/master/site/dat/wiki/AutoStop.wiki

If you want to make this really JMeter-style feature, you should introduce a UI input field where user inputs the name of environment variable to set. That would need a but more programming. However, I won't insist on this.

ricardasomni:
> For this to complete, you'd need to also document this feature, so other people would be able to discover it, too. It lives in this file: https://github.com/undera/jmeter-plugins/blob/master/site/dat/wiki/AutoStop.wiki
> 
> If you want to make this really JMeter-style feature, you should introduce a UI input field where user inputs the name of environment variable to set. That would need a but more programming. However, I won't insist on this.

Thanks, wiki updated. I don't have much experience in developing JMeter plugins, but this feature I believe could be interesting for many. The use case here is to use JMeter in CI envirnoment to run fully automated performance test after deployment and stop test + alert based on response times. Feel free to suggest other changes or update if you see value in this.
Only other way to fail test and take actions based on average response/latency times is via Taurus, but that's a completely different realm.

undera:
This plugin is so rarely changed, that I'm happy to see any improvement. 
Thanks for contributing!

PR #596, opened 2023-06-24
Version 4.0 add command FIND a line in the filename

vdaburon:
Command FIND with FIND_MODE : a string with SUBSTRING (Default) or EQUALS and a regular expression with REGEX_FIND (contains) and REGEX_MATCH (entire region the pattern)


undera:
I'll do release into Maven, then will re-start the GH Action.
Ideally, we should have separate PR for changing the descriptor, for better coordination

PR #597, opened 2023-07-07
add new tool vdn-junit-reporter-kpi-from-jmeter-dashboard-stats

vdaburon:
Add tool.
This tool creates a JUnit XML file with KPI rules from JMeter JMeter Dashboard stats.json and generates a result file in JUnit XML format and other formats  : html, csv and json.


undera:
Very nice plugin!

PR #604, opened 2023-07-22
Update of WebDriver to Selenium 4.10.0

rbourga:


undera:
One of the libraries fails to download, @rbourga please check it out

PR #610, opened 2023-08-21
BlazeMeter Repositories Core 1.0

3dgiordano:
New plugin for Correlation Recorder

### DIFF
--- a/site/dat/repo/blazemeter.json
+++ b/site/dat/repo/blazemeter.json
@@ -1214,11 +1214,31 @@
           "jackson-core>=2.13.0" : "https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/jackson-core-2.13.0.jar",
           "jackson-module-jaxb-annotations>=2.10.2":"https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/jackson-module-jaxb-annotations-2.10.2.jar",
 		  "jmeter-bzm-commons>=0.2.1":"https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/jmeter-bzm-commons-0.2.1.jar",
-		  "json>=20190722":"https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/json-20190722.jar",
-		  "bzm-repositories-plugin>=1.0":"https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/bzm-repositories-plugin-1.0.jar"
-        }
+		  "json>=20190722":"https://github.com/Blazemeter/CorrelationRecorder/releases/download/v2.1/json-20190722.jar"
+        },
+		"depends": [
+			"bzm-repositories"
+		]
       }
     }
+  },
+  {
+    "id": "bzm-repositories",
+    "name": "BlazeMeter - Repositories Core",
+    "description": "BlazeMeter repositories functionality for the Correlation Recorder Plugin",
+    "screenshotUrl": "",
+    "helpUrl": "",
+    "vendor": "BlazeMeter",
+    "markerClass": "com.blazemeter.jmeter.repository.plugin.system.SystemManager",
+    "componentClasses": [],
+    "versions": {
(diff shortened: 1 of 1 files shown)


undera:
Does it create circular dependency between the two plugins?

3dgiordano:
Hi @undera 

Yes.
does that cause some kind of problem?  I found no problems when I tried it with Plugin manager.

At first Correlation Recorder will promote the new plugin (Blazemeter repositories) and the new plugin depends on Correlation Recorder plugin.

If this can cause a problem, I can remove the dependency to CR plugin from the Blazemeter Repositories and add in the description that it depends on that plugin.

3dgiordano:
Hi @undera 

Yes.
does that cause some kind of problem?  I found no problems when I tried it with Plugin manager.

At first Correlation Recorder will promote the new plugin (Blazemeter repositories) and the new plugin depends on Correlation Recorder plugin.

If this can cause a problem, I can remove the dependency to CR plugin from the Blazemeter Repositories and add in the description that it depends on that plugin.

undera:
Circular dependencies are usually bad idea. You need to choose which of the plugins truly depend on another. Otherwise this will cause trouble in any place that resolves dependencies. Pmgr itself has special protection from it, but I still would like to be safe.

3dgiordano:
Hi @undera 

Ok, I removed the circular dependency.
I hope now everything is fine now for your approval and merge.

Thank you.

undera:
Why "screenshotUrl", helpUrl" and "componentClasses" are empty?

3dgiordano:
Hi @undera 
They are not required at this time.
I took the same reference structure from other "Core" components already existing in the catalog.

When the tasks related to documentation and public release are done (It's not something on my side), then I will submit the change to screenshot and the link to help url.
The component mechanism does not apply, only markerClass applies.

undera:
We should not offer for users the plugins that have no reference to documentation explaining what is that. 
JMeter plugin items should carry JMeter plugins insight them, that's the idea. JMeter Core is the only exception, by obvious reasons.
If this component is not JMeter plugin at all - then it's not correct to distribute it via jmeter-plugins.org. Maybe it should just be bundled with the actual plugin it relates to.

3dgiordano:
Hi @undera 
To clarify a bit, this is the first plugin for the Correlation Recorder plugin, and it relies on the extension loading mechanism in lib/ext, and will use the Plugin Manager to allow installation, versioning and updates.
These plugins will be able to extend jmeter as well.
It doesn't use the jmeter component mechanism, but it does use markerClass to detect if the extension is installed in lib/ext.
There are no plans to do a full blown release of this extension system at this time, so at this time the metadata is discrete.

But if there will be documentation and launch of this product soon.
When this is done, the information will be incorporated into the plugin, for now we need it to be there, but we don't want to present it for now.
BlazeMeter will be rolling this out soon.

undera:
"for now we need it to be there" - who is "we"? 
I'm trying to guard users from strange things that they are unable to understand or read about. AFAIK users don't "need it to be there". With my respect to commercial companies - they should not bypass basic rules of the project. Let's wait when the documentation is released and then do proper release. 
Also, if you say there's no plan for full blown separate release - just incorporate it into Correlation Recorder, where it belongs to.

3dgiordano:
Ok. This plugin system for our open source Correlstion Recorder plugin allows third party functionality to be incorporated that is not part of the project code.  That is the system that will not be officially released to the public now. And because this new extension plugin not is part of the Correlation Recorder, that dependency need to by released separately as a new Plugin. Plugin Manager and the system of distribution it's the correct place to add lib/ext extensions to Jmeter.

If having a screenshot and a help url is what's needed, I'll add it. I didn't think it was required for initial record creation. 
I see to incorporate it at soon as possible in order to continue.

3dgiordano:
Hi @undera 

I already added the link to the help documentation and the screenshot.
I also removed the "Core" concept that can confuse users.

Thanks for the suggestions.

undera:
Thanks, it got much better!

PR #617, opened 2023-09-28
Addition of new plugins for Outlier Detection, Apdex and Results Comparison.

rbourga:
Users will be prompted to unistall the deprecated plugins and install the new ones.
Will remove the old plugins from the list by end of year 2023.


undera:
I suggest adding to deprecated plugin's names `(deprecated)` suffix, same way it's done to "Oauth Sampler". It will be better visible to users.

PR #624, opened 2023-11-07
Update of WebDriver to Selenium 4.13.0.

rbourga:
Removal of obsolete WebDriver versions to lighten the file.
Addition of new entry for WebDriver 4.13.0.
This is the last update of WebDriver until JMeter upgrades to a version of Java greater than or equal to 11.


site/dat/repo/jpgc-plugins.json
```diff
@@ -748,187 +748,6 @@
       "com.googlecode.jmeter.plugins.webdriver.sampler.gui.WebDriverSamplerGui"
     ],
     "versions": {
-      "1.4.0": {
```
undera, line 751: According to usage statistics, there's still many people using older versions, thus these versions should be still offered.  The only versions I we might delete are those that are old _and_ have less than 1% of usage. There are just couple of those.

undera (changes requested):
Please restore deleted section. Maybe only couple of entries can be removed.

undera:
One of the libs fails to download

PR #626, opened 2023-11-23
Version 5.0 add READMULTI command

vdaburon:
STS version 5.0 add READMULTI command and correct bug with SAVE command ADD_TIMESTAMP=true


undera:
Published to maven

PR #629, opened 2023-12-08
[fix] - JedisException when test ends 

v-barros:
Fixing the Exception:  redis.clients.jedis.exceptions.JedisException: Could not return the broken resource to the pool
Which is triggered when jmeter calls testEnded() method. 
Evidence below:
image

I added a log to show the total number of idle connections before calling ``pool.destroy()`` at ``testEnded(String host)`` and noticed that the number of idle connections was equal to the number of exceptions. As we can see here:
image

Then I forced JMeter to consume all the connections before calling ``pool.destroy()``.
I know the problem is in Jedis itself, as it was already addressed in https://github.com/redis/jedis/pull/3353. The other possible solution would be updating the Jedis version, but I don't know if that would lead to any compatibility issue.


joseph-roy:
This fix is not with the .06 version of the plugin. Can you share the latest artefact location?

undera:
> This fix is not with the 0.6 version (built much earlier than this fix) of the plugin. Can you share the latest artefact location?

I have just released the new version.

PR #635, opened 2024-02-22
add initial version of HAR Importer

besessener:
Add a new importer that is able to take the browsers HAR files (HTTP Archive) and convert them into HTTP Samplers, Constant Timers, add cookies and headers and sort them into modules.

The motivation is to allow customers to record http requests without any knowledge of JMeter and without the hurdle of proxies.

@see https://github.com/Qytera-Gmbh/JMeterHARImporterPlugin/blob/main/README.md

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -2008,5 +2008,23 @@
 	        "depends":["jmeter-core"]
 		}
 	}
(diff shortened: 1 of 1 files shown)


undera:
Congrats on the new plugin!

PR #637, opened 2024-03-04
Parallel Sampler 0.12

jloisel:
Fix Parallel Sampler to support JMeter 5.6.x, with backward compatibility with older versions of JMeter.

### DIFF
--- a/site/dat/repo/blazemeter.json
+++ b/site/dat/repo/blazemeter.json
@@ -266,6 +266,16 @@
         "depends": [
           "jmeter-http"
         ]
+      },
+      "0.12": {
+        "changes": "Fix parallel controller for JMeter 5.6+",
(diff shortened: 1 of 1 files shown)


undera:
Sorry, but you can't hijack somebody's plugin with your custom releases. Please contact the original plugin authors and contribute your change there

jloisel:
Hi, 

Aren't you the main maintainer of jmeter bzm plugins?

https://github.com/Blazemeter/jmeter-bzm-plugins/graphs/contributors

I understand, but the parallel sampler plugin isn't maintained anymore, I have opened a pull request:
https://github.com/Blazemeter/jmeter-bzm-plugins/pull/94

Nobody answered since July 2023. 

What else can we do?

undera:
I don't work for BlazeMeter for many years, and I'm not  the maintainer of that plugin.

Did you try sending email to their support? I their org profile I see other repos are getting updated. Maybe you need to find the right person. You may want to ask for maintainer role even.

PR #641, opened 2024-03-18
vdn-har-convertor-jmeter-tool v3.0

vdaburon:
vdn-har-convertor-jmeter-tool v3.0.
Add Load generated script if no error. File Chooser select only file and no directory


vdaburon:
Thank Andrey for your reactivity.

PR #648, opened 2024-03-30
vdn-har-convertor-jmeter-tool v3.2

vdaburon:
vdn-har-convertor-jmeter-tool v3.1, use library har-to-jmeter-convertor-2.2 that correct bugs.
* Remove the header 'Content-length' because the length is computed by JMeter when the request is created. 
* POST or PUT could have query string and body with content so add query string to the path. 
* Set Content Encoding to UFT-8 for POST or PUT method and request Content-Type : application/json. 
* Add body data content in Record.xml for PUT and PATCH methods.


vdaburon:
Sorry i find a bug for POST with Query String and body JSON. 

**Don't push this version**

vdaburon:
OK vdn-har-convertor-jmeter-tool version 3.2

vdaburon:
Thanks Andrey

PR #656, opened 2024-05-27
Fix news/changes on release v2.5

Baraujo25:
There was a mistake while copying and pasting last release as template. Changes for version 2.5 weren't updated.


Baraujo25:
@undera Pipe hanged, I'm unable to stop it or restart it. Any thoughts?

PR #690, opened 2025-01-21
Custom Thread Groups - "Same user on each iteration" support for JMeter 5.2 onwards

3dgiordano:
Custom Thread Groups - "Same user on each iteration" support for JMeter 5.2 onwards

In 2019, support for the concept of "Same user on each iteration" was added to JMeter 5.2.
The "Thread Groups" of the "Custom Thread Groups" plugin need to support this concept.
The proposal to support it is to raise the minimum supported version to JMeter to 5.2 and change the plugin to add the support and change version to 3.0 (because no longer support JMeter prior to 5.2).

Unfortunately, when the functionality was implemented in JMeter 5.2, it was done through a modification in its JMeterThread constructor and a private internal flag, which brings difficulties for its support from the current JMeter 2.13 support (and probably could not be solved by reflection even if it were attempted using that method)

According to the statistics provided by jmeter-plugins, more than 80% of the active JMeter are equal to or higher than JMeter 5.2 (released at the end of 2019, more than 5 years ago).
In case users want to use the plugin in versions prior to JMeter 5.2, they must use the previous version 2.10.

I hope that this PR proposal is welcomed to begin updating the support for new JMeters.

PS: I tried to modify as little as possible to make the review as easy as possible.
I also discovered a bug in jmeter-plugins-emulators regarding how it assigns the properties of where the save services file is located, which caused the ss.props file to not be found in new jmeters. The WA solve that problem.

### DIFF
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
@@ -29,7 +29,7 @@ public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements Pa
         add("Time Unit: ", groupPanel);
 
         add("Thread Iterations Limit: ", iterations);
-        add("Same user on each iteration", sameUserBox);
+        add("Same User On Each Iteration", sameUserBox);
         add("Log Threads Status into File: ", logFile);
 
         if (showConcurrencyLimit) {
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
@@ -12,6 +12,7 @@ import java.util.Enumeration;
 public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements ParamsPanel {
     protected JTextField logFile = new JTextField();
     protected JTextField iterations = new JTextField();
+    private final JCheckBox sameUserBox = new JCheckBox();
     protected JTextField concurrLimit = new JTextField();
     protected ButtonGroup unitGroup = new ButtonGroup();
     protected JRadioButton unitSeconds = new JRadioButton("seconds");
@@ -28,6 +29,7 @@ public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements Pa
         add("Time Unit: ", groupPanel);
 
         add("Thread Iterations Limit: ", iterations);
+        add("Same user on each iteration", sameUserBox);
         add("Log Threads Status into File: ", logFile);
 
         if (showConcurrencyLimit) {
@@ -38,6 +40,7 @@ public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements Pa
     public void modelToUI(AbstractDynamicThreadGroup tg) {
         logFile.setText(tg.getLogFilename());
         iterations.setText(tg.getIterationsLimit());
+        sameUserBox.setSelected(tg.getSameUser());
         concurrLimit.setText("1000");
         unitMinutes.setSelected(true);
         if (tg instanceof ArrivalsThreadGroup) {
@@ -57,6 +60,7 @@ public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements Pa
     public void UItoModel(AbstractDynamicThreadGroup tg, JMeterVariableEvaluator evaluator) {
         tg.setLogFilename(evaluator.evaluate(logFile.getText()));
         tg.setIterationsLimit(evaluator.evaluate(iterations.getText()));
+        tg.setSameUser(sameUserBox.isSelected());
         if (unitGroup.getSelection() != null) {
             tg.setUnit(unitGroup.getSelection().getActionCommand());
         }
@@ -70,6 +74,7 @@ public class AdditionalFieldsPanel extends ArrangedLabelFieldPanel implements Pa
     public void clearUI() {
         logFile.setText("");
         iterations.setText("");
+        sameUserBox.setSelected(true);
         concurrLimit.setText("1000");
         unitMinutes.setSelected(true);
     }
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
@@ -38,6 +38,11 @@ public class ArrivalsThreadGroup extends AbstractDynamicThreadGroup {
         }
     }
 
+    @Override
+    public JMeterThread addNewThread(int i, StandardJMeterEngine standardJMeterEngine) {
+        return null;
+    }
+
     public int getNumThreads() {
         return threads.size();
     }
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
@@ -38,11 +38,6 @@ public class ArrivalsThreadGroup extends AbstractDynamicThreadGroup {
         }
     }
 
-    @Override
-    public JMeterThread addNewThread(int i, StandardJMeterEngine standardJMeterEngine) {
-        return null;
-    }
-
     public int getNumThreads() {
         return threads.size();
     }
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
@@ -53,6 +53,11 @@ public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
     }
 
     @Override
+    public JMeterThread addNewThread(int i, StandardJMeterEngine standardJMeterEngine) {
+        return null;
+    }
+
+    @Override
     public boolean isLimitReached() {
         return !running || !threadStarter.isAlive();
     }
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
@@ -52,10 +52,6 @@ public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
         return (int) Math.round(getTargetLevelAsDouble());
     }
 
-    @Override
-    public JMeterThread addNewThread(int i, StandardJMeterEngine standardJMeterEngine) {
-        return null;
-    }
 
     @Override
     public boolean isLimitReached() {
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
@@ -52,7 +52,6 @@ public class ConcurrencyThreadGroup extends AbstractDynamicThreadGroup {
         return (int) Math.round(getTargetLevelAsDouble());
     }
 
-
     @Override
     public boolean isLimitReached() {
         return !running || !threadStarter.isAlive();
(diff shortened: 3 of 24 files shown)


plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
```diff
@@ -28,6 +29,7 @@ public AdditionalFieldsPanel(boolean showConcurrencyLimit) {
         add("Time Unit: ", groupPanel);
 
         add("Thread Iterations Limit: ", iterations);
+        add("Same user on each iteration", sameUserBox);
```
undera, line 32: Please use the same work capitalization style as on other labels

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
```diff
@@ -70,6 +74,7 @@ public void UItoModel(AbstractDynamicThreadGroup tg, JMeterVariableEvaluator eva
     public void clearUI() {
         logFile.setText("");
         iterations.setText("");
+        sameUserBox.setSelected(true);
```
undera, line 77: It it really the default behavior in JMeter to enable it?

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
```diff
@@ -38,6 +38,11 @@ public void start(int groupIndex, ListenerNotifier listenerNotifier, ListedHashT
         }
     }
 
+    @Override
```
undera, line 41: All of these overrides with `null` result look strange. Can you shortly explain why is it like that?

undera (changes requested):
Please address my comments

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
3dgiordano: Good catch, I copied the original text from JMeter and didn't realize it. Thanks.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AdditionalFieldsPanel.java
3dgiordano: JMeter TheadGroup has it enabled by default.
https://github.com/apache/jmeter/blob/ab65ec18103fad25cbd85aa05078bb7f89c5caca/src/core/src/main/kotlin/org/apache/jmeter/threads/AbstractThreadGroupSchema.kt#L44

These clearUI use the default values ​​but the model always wins.
What you made me see is that I do not actually have the default value set to true in the model when the property does not exist, now I replicated the same as JMeter's ThreadGroup.
Thanks for the question, I'm looking to make that fix to imitate the same behavior as JMeter's ThreadGroup.

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroup.java
3dgiordano: addNewThread exist in the Abstract of ThreadGroup but never was used (is not called by any class). It's like a "garbage".
https://github.com/search?q=repo%3Aapache%2Fjmeter%20addNewThread&type=code

OpenModelThreadGroup by example add the following comment "Will not be implemented as the semantics of the API is unclear"
I'll add a comment similar to the one in OpenModel to make it clear why.

undera: Maybe implement it in AbstractDynamicThreadGroup, so the inheritors don't need to bother?

plugins/casutg/src/test/java/kg/apc/jmeter/threads/AbstractSimpleThreadGroupImpl.java
```diff
@@ -13,4 +14,9 @@ public void scheduleThread(JMeterThread thread, long now) {
     public Map<JMeterThread, Thread> getAllThreads() {
         return this.allThreads;
     }
+
+    @Override
```
undera, line 18: Same here - can AbstractSimpleThreadGroup bother implementing this, so the implementation classes don't have to?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroupGuiTest.java
```diff
@@ -15,6 +16,9 @@ public class ArrivalsThreadGroupGuiTest {
     @BeforeClass
     public static void setUpClass() throws IOException {
         TestJMeterUtils.createJmeterEnv();
+        // Workaround for new JMeters
```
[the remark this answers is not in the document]
3dgiordano: It's "external" to this plugin.
I think it is not harmful.
But if you think it's okay, I'll do the fix in the emulator as well.
I suppose the tests of the other plugins will tell me if it affects the JMeter 2.13 runtime or not.
Let's try it and see.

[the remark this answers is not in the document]
undera: Yes, if no harm - let's change the emulators

plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroup.java
```diff
@@ -52,6 +52,7 @@ public int getNumThreads() {
         return (int) Math.round(getTargetLevelAsDouble());
     }
 
+
```
undera: Unwanted change?

plugins/casutg/src/main/java/kg/apc/jmeter/threads/SteppingThreadGroup.java
```diff
@@ -1,5 +1,6 @@
 package kg.apc.jmeter.threads;
 
+import org.apache.jmeter.engine.StandardJMeterEngine;
```
undera: Unwanted change?

plugins/casutg/src/main/java/kg/apc/jmeter/threads/UltimateThreadGroup.java
```diff
@@ -1,6 +1,7 @@
 package kg.apc.jmeter.threads;
 
 import kg.apc.jmeter.JMeterPluginsUtils;
+import org.apache.jmeter.engine.StandardJMeterEngine;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/control/VirtualUserControllerTest.java
```diff
@@ -4,6 +4,7 @@
 import com.blazemeter.jmeter.threads.arrivals.ArrivalsThreadGroupTest;
 import kg.apc.emulators.TestJMeterUtils;
 import org.apache.jmeter.sampler.DebugSampler;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroupGuiTest.java
```diff
@@ -2,6 +2,7 @@
 
 import kg.apc.emulators.TestJMeterUtils;
 import org.apache.jmeter.testelement.TestElement;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/arrivals/ArrivalsThreadGroupTest.java
```diff
@@ -3,6 +3,7 @@
 import kg.apc.emulators.TestJMeterUtils;
 import org.apache.jmeter.sampler.TestAction;
 import org.apache.jmeter.threads.AbstractThreadGroup;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/arrivals/FreeFormArrivalsThreadGroupGuiTest.java
```diff
@@ -1,6 +1,7 @@
 package com.blazemeter.jmeter.threads.arrivals;
 
 import kg.apc.emulators.TestJMeterUtils;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/arrivals/FreeFormArrivalsThreadStarterTest.java
```diff
@@ -5,6 +5,7 @@
 import org.apache.jmeter.testelement.property.CollectionProperty;
 import org.apache.jmeter.testelement.property.StringProperty;
 import org.apache.jmeter.threads.ListenerNotifier;
+import org.apache.jmeter.util.JMeterUtils;
```
undera, line 8: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadGroupTest.java
```diff
@@ -25,6 +25,7 @@
 import org.apache.jmeter.threads.ListenerNotifier;
 import org.apache.jmeter.threads.TestCompiler;
 import org.apache.jmeter.timers.ConstantTimer;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

plugins/casutg/src/test/java/com/blazemeter/jmeter/threads/concurrency/ConcurrencyThreadStarterTest.java
```diff
@@ -2,6 +2,7 @@
 
 import com.blazemeter.jmeter.threads.arrivals.ArrivalsThreadGroupTest;
 import kg.apc.emulators.TestJMeterUtils;
+import org.apache.jmeter.util.JMeterUtils;
```
undera: Unwanted change?

undera (changes requested):
Please remove all the unwanted changes

plugins/casutg/src/test/java/kg/apc/jmeter/threads/AbstractSimpleThreadGroupImpl.java
```diff
@@ -13,4 +13,5 @@ public void scheduleThread(JMeterThread thread, long now) {
     public Map<JMeterThread, Thread> getAllThreads() {
         return this.allThreads;
     }
+
```
undera, line 16: Unwanted change

3dgiordano:
Thanks @undera for all the help

You are very quick at reviewing. I saw all the "Unwanted changes" with the commit of the requested changes. 
They have already been resolved and fixed.

The PR is much smaller. 

Thank you very much for the suggestions and for allowing the emulator to be updated.

As you may have noticed I uploaded a version of the emulator. I noticed most of the plugins are at 0.2 so it will only be known if it affects or not if they are updated to 0.5.

IMPORTANT!
Making the change in the model and testing the plugin I realized that it seems that "The default logic" does not work in this group of plugins (an unreported issue).

Please don't approve the PR, I need to review this issue and see if I can resolve this issue as well on this PR (or maybe we leave it for someone else, you tell me what do you think about)

undera:
Looks good now, thank you very much
Please lead the next step of updating the plugin descriptor

3dgiordano:
Thanks @undera 

I will proceed with the plugin descriptor as soon as the plugin is available.

Thanks.

undera:
OOps. I did not see your message when merged.
Ok, not a catastrophe until we release the descriptor

3dgiordano:
@undera Don't worry.
I need to create an Issue and see if I have time to solve that in another PR.
Not is related with the actual PR, and not is good idea to integrate on this. 
The issue has been around for a while and no one has reported it, I guess it's not that critical.

As soon as I can, I'll create the update for the descriptors.
I guess we have to wait for the publish to maven to be resolved and for them to be available for approval.

We'll get in sync with the next PR of the plugin release.

Thank you very much.

PR #691, opened 2025-01-22
Custom Thread Groups - Allow to run on JMeter lower than 5.2

3dgiordano:
The latest change made in Custom Thread Groups version 3.0 allows the new JMeterThread from JMeter 5.2 to be used in ConcurrencyThreadGroup, ArrivalsThreadGroup and FreeFormArrivalsThreadGroup, but breaks compatibility with JMeter older than 5.2.

I am making a proposal to allow the use of JMeters older than 5.2 using the old JMeterThread constructor.
This applies only to ConcurrencyThreadGroup, ArrivalsThreadGroup and FreeFormArrivalsThreadGroup.

Apologies for not realizing earlier that it was possible to incorporate this possibility, I later realized that it would be easy to incorporate it and thus surely not break the execution of those who still use JMeter lower than 5.2.

I can incorporate this "version checking" and disable the new field in the UI if necessary, although it is adding code for probably a small group of users.

I look forward to hearing what you think about this proposal.

### DIFF
--- a/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
+++ b/plugins/casutg/src/main/java/com/blazemeter/jmeter/threads/AbstractThreadStarter.java
@@ -16,6 +16,10 @@ import org.slf4j.Logger;
 
 public abstract class AbstractThreadStarter extends Thread {
     private static final Logger log = LoggerFactory.getLogger(AbstractThreadStarter.class);
+    private static final String[] jmeterVersion =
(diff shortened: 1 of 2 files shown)


undera:
If the majority of users is using 5.2+, then it's not worth the complication. Users can install the older version as a workaround

3dgiordano:
@undera OK.

I'm going to close the PR, as it doesn't seem to be important.
Thanks again

PS: I noticed that the build for version 3.0 is already in maven, I'm going to proceed to create the new PR with the description of the new version and update the wiki to indicate the compatibility information with JMeter versions.

PR #693, opened 2025-02-06
Version 6.1 for har-convertor-jmeter-tool

vdaburon:
Version 6.1 Use new library har-to-jmeter-convertor 6.1 to correct a NullPointerException when creating the Recording XML file.


vdaburon:
Thanks Andrey

PR #702, opened 2025-03-14
Bump/feather wand 1.0.3

QAInsights:
Added supported for most of the JMeter elements


undera:
It would be nice if you include short comment of what has changed in each version. There is a special field for that called "changes". For example the other contributor: https://github.com/undera/jmeter-plugins/commit/62c639933c8174fb381aeec5275327fef640956c

QAInsights:
Thanks @undera will do that in the next version.

PR #817, opened 2026-04-08
Add perfsage SLO reporting listener to the plugin registry

aashu320:
Adds community plugin **perfsage-slo-reporter** to the catalog.

- **Repo:** https://github.com/perfsage/perfsage-slo-reporter
- **Release / JAR:** https://github.com/perfsage/perfsage-slo-reporter/releases/tag/v0.1.0
- **License:** MIT
- Descriptor matches [Plugin Repository Descriptor Format](https://jmeter-plugins.org/wiki/PluginRepositoryDescriptorFormat/).

`downloadUrl` points at the GitHub Release asset; `screenshotUrl` uses `raw.githubusercontent.com` for a stable image URL.


undera:
This plugin ships nothing that is not available in existing JMeter and plugins. Consider joining efforts into existing projects than increasing the fragmentation.

That said, your plugin has good instruction on how to add it into JMeter if anyone wants exactly that.

aashu320:
@undera I am a performance engineer and I was actually blocked while developing such scenario using jmeter. AFAIK, there is no native functionality or plugin in jmeter that provides comparison and analysis of results for a specific SLO. 

There are tools like Taurus (wrapper to jmeter), which actually fulfil this requirement but **no native functionality as I said**. If you are aware about any such plugin, please do let me know. It will be really helpful.

undera:
Jmeter has Assertion concept that addresses what your plugin does.

aashu320:
@undera **No, it doesn't.** JMeter assertion is specific to 'sampler' or a specific request. Although we can achieve similar behaviour - it won't be an efficient way of doing it. JMeter assertions need to run after each request, it will impact jmeter performance itself.

I already tried to use samplers, but didn't look like a good pluggable solution for any performance pipeline. Taurus is another solution which is like my plugin, but it adds another layer of execution. Even before starting on this, I checked all available jmeter plugins and functionality. I wanted to develop something which is jmeter specific, and fills this specific gap.

undera:
Fair enough, let's give it a try. Please resolve the conflicts with base file. And also `SLOReporter` is plain class and not JMeter component, useless for descriptor, should be removed.

aashu320:
@undera  Done. Please check.

undera:
It has failed CI - take a look

undera (changes requested):
Be modest. You descriptor should be last in the list, not the first. 
And please pull from master, I fixed the CI issue there.

undera:
Move it to the end and I'll merge

PR #821, opened 2026-04-10
Update JAAR plugin registry — fix package paths, add v6.0.0

sagaraggarwal86:
The marker class and component class paths were renamed to io.github.sagaraggarwal86.jmeter

- Updated plugin description to reflect features added since last registry update:
  - TPS SLA threshold
  - Data-only HTML report (**no AI required**)
  - CLI fallback to data-only report on AI failure


undera:
You need to resolve the conflict

PR #834, opened 2026-05-09
NEW: README Sampler

QAInsights:


site/dat/repo/various.json
```diff
@@ -3368,5 +3368,23 @@
         "depends": []
       }
     }
+  },
+  {
+    "id": "jmeter-readme-sampler",
+    "name": "README Sampler",
+    "description": "A JMeter sampler that lets you embed Markdown documentation directly inside your test plan with a live GitHub-style preview and deep-link navigation to tree nodes.",
```
undera: I'd argue that sampler is poor choice for this type of component, because sampler can't be attached to other components as a contextual help piece, and has placement restrictions.
I would suggest to use "Config" element as this type of documenting component - it is closer by intuition of it, and can be placed more freely across Test Plans.
WDYT?

QAInsights: Thanks for checking out. Placement is one and I went by performance intuition. 

That is why I intentionally [disabled it,](https://github.com/QAInsights/jmeter-readme-sampler/blob/28a0e556b99ab2c81e6a0bdba575a79ff7f63875/src/main/java/com/qainsights/jmeter/readme/readme/ReadMeSampler.java#L17) it cannot be enabled by the user after adding it.  

It will be always disabled; will not influence any threads. 

**Also the main rationale of choosing Sampler is by following Dummy Sampler plugin. :)** 

Please share your views. 

Let me also try to refactor this. 

Disabled

QAInsights: @undera I refactored it to Config element, testing it now, will push a new commit soon. Thanks.

PR #850, opened 2026-06-16
Add HAR vs JMeter Listener plugin

Vishhh-alt:
New plugin submission: HAR vs JMeter Listener

Compares real browser load time (HAR files) with JMeter sampler 
results side by side. Native JMeter Listener, installs via lib/ext.

GitHub: https://github.com/Vishhh-alt/HAR_vs_Jmeter
JAR: https://github.com/Vishhh-alt/HAR_vs_Jmeter/releases
Author: Visvashwarr Venugopal
Requires: JMeter 5.x, Java 11+, Python 3.7+

### DIFF
--- a/site/dat/repo/various.json
+++ b/site/dat/repo/various.json
@@ -3468,5 +3468,19 @@
         }
       }
     }
+  },
+  {
+  "id": "har-jmeter-listener",
+  "name": "HAR vs JMeter Listener",
+  "description": "Compares real browser load time (HAR files) with JMeter sampler results side by side. Matches requests by URL, generates an HTML report with diffs, waterfall view, and a built-in Why-the-diff reference guide. Works in Live mode and Offline mode.",
+  "helpUrl": "https://github.com/Vishhh-alt/HAR_vs_Jmeter",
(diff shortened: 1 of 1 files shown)


undera:
I don't agree this being published into catalogue. It has very little need to be a plugin, it is in fact a Python script to build a report. The use-case is not live JMeter test, it's all "after the test report building". The real use case is "save JMeter JTL report then run it together with HAR via Python script". Moreover, running Python script from inside Java would be very fragile and hard to setup properly.

Vishhh-alt:
Thank you for the honest and detailed feedback, Andrey — I appreciate you taking the time to review this.

You're right on all counts. The core value is the Python report engine, and the Java wrapper adds fragility (Python path setup, temp file extraction, OS differences) without meaningful benefit. The real use case is post-test: save the JTL/Summary CSV, capture the HAR, run the Python script, get the report.

I'll close this PR and instead publish the standalone Python script as an open source tool on GitHub, which is a cleaner fit for the actual workflow.

Thanks again for the direction — this is exactly the kind of feedback that makes a better tool.

PR #880, opened 2026-09-01
Add Adaptive Capacity Listener plugin descriptor

hilalsidhic:
### Description

Adds the descriptor for **Adaptive Capacity Listener** (`adaptive-capacity-listener`) to `site/dat/repo/various.json`.

**Adaptive Capacity Listener** is an Apache JMeter listener and backend monitor plugin designed for adaptive capacity and resilience testing. It continuously tracks response performance over configurable evaluation windows and automatically halts tests when configured degradation or SLA thresholds are breached, injecting a synthetic failure sample to clearly mark threshold breaches in summary reports and HTML dashboards.

### Features
- **Multi-Policy Degradation Evaluation**:
  - `PERCENT_LATENCY`: Detects abnormal percentile latency growth compared to prior evaluation stages.
  - `ABSOLUTE_LATENCY`: Enforces a hard latency ceiling in milliseconds.
  - `ERROR_RATE`: Monitors error percentage per evaluation window.
  - `ERROR_COUNT`: Enforces a maximum acceptable error count per window.
- **Headless & CLI Compatible**: Operates both in JMeter GUI mode (under **Add → Listener**) and in headless/CLI mode via Backend Listener (`AdaptiveCapacityBackendListener`).

### Links
- **Repository & Docs**: https://github.com/hilalsidhic/AdaptiveCapacityJmeterPlugin
- **Release (v1.0.1)**: https://github.com/hilalsidhic/AdaptiveCapacityJmeterPlugin/releases/tag/v1.0.1
- **Download URL**: https://github.com/hilalsidhic/AdaptiveCapacityJmeterPlugin/releases/download/v1.0.1/AdaptiveCapacityJmeterPlugin-1.0.1.jar

PR #883, opened 2026-09-02
Add relative percentile degradation and error count ceiling criteria to AutoStop listener

hilalsidhic:
### Description

As discussed in #880, this PR adds the missing criteria to the **Auto-Stop Listener (`jpgc-autostop`)**:

1. **Relative Window-to-Window Percentile Degradation**:
   - Compares the percentile latency of the current observation window ($P_n$) against the baseline percentile of the previous window.
   - Automatically shuts down the test if the relative growth percentage exceeds the configured threshold.
   - Fields: Percentile rank (e.g. `90`), max growth `%` (e.g. `50%`), observation window in seconds (e.g. `5s`).

2. **Per-Window Absolute Error Count Ceiling**:
   - Automatically shuts down the test if the absolute count of failed samples within a duration window exceeds a set threshold.
   - Fields: Max allowed error count, window duration in seconds.

3. **GUI Checkboxes & Layout**:
   - Added checkboxes to the left of all 6 criteria to allow users to selectively enable/disable individual conditions without deleting their values.
   - Standardized spacing and alignment across all condition rows.

4. **Input Safety & Error Handling**:
   - Fixed latent `NumberFormatException` when fields are left blank or unconfigured.

5. **Examples & Unit Tests**:
   - Added 11 unit tests in `AutoStopTest` covering threshold breaches, window resets, and disabled states.
   - Added two sample `.jmx` test plans under `examples/`:
     - `AutoStopExample_Percentile.jmx`
     - `AutoStopExample_ErrCount.jmx`

### Verification
- `mvn clean test package` runs with 0 failures, 0 errors across all 40 tests.
- Verified in JMeter GUI with live execution and chart rendering.

### DIFF
--- a/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
+++ b/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
@@ -32,6 +32,7 @@ import java.net.DatagramPacket;
 import java.net.DatagramSocket;
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class AutoStop
         extends AbstractListenerElement
@@ -83,13 +84,14 @@ public class AutoStop
     private int testValueCustomSec = 180;
     private boolean skipIter = false;
 
+    private final Object relWindowLock = new Object();
     private SamplingStatCalculator relWindowStatCalc = new SamplingStatCalculator();
     private long relWindowStart = 0;
     private double previousWindowPercentile = -1;
     private float testValueRelPercentileValue = 0;
     private int testValueRelWindowSecs = 0;
     private float testValueRelThresholdPct = 0;
-    private int errorCountInWindow = 0;
+    private final AtomicInteger errorCountInWindow = new AtomicInteger(0);
     private long errorCountWindowStart = 0;
     private int testValueErrorCount = 0;
     private int testValueErrorCountSec = 0;
@@ -103,12 +105,14 @@ public class AutoStop
     public void sampleOccurred(SampleEvent se) {
         long sec = System.currentTimeMillis() / 1000;
 
-        if (testValueRelWindowSecs > 0) {
-            relWindowStatCalc.addSample(se.getResult());
+        if (testValueRelPercentileValue > 0 && testValueRelWindowSecs > 0 && testValueRelThresholdPct > 0) {
+            synchronized (relWindowLock) {
+                relWindowStatCalc.addSample(se.getResult());
+            }
         }
 
-        if (testValueErrorCount > 0 && !se.getResult().isSuccessful()) {
-            errorCountInWindow++;
+        if (testValueErrorCount > 0 && testValueErrorCountSec > 0 && !se.getResult().isSuccessful()) {
+            errorCountInWindow.incrementAndGet();
         }
 
         if (curSec != sec) {
@@ -171,10 +175,14 @@ public class AutoStop
             }
 
             // Relative window-to-window percentile degradation check
-            if (testValueRelWindowSecs > 0 && testValueRelThresholdPct > 0) {
+            if (testValueRelPercentileValue > 0 && testValueRelWindowSecs > 0 && testValueRelThresholdPct > 0) {
                 if (sec - relWindowStart >= testValueRelWindowSecs) {
-                    double currentPct = relWindowStatCalc.getPercentPoint(testValueRelPercentileValue).doubleValue();
-                    if (previousWindowPercentile >= 0 && previousWindowPercentile > 0) {
+                    double currentPct;
+                    synchronized (relWindowLock) {
+                        currentPct = relWindowStatCalc.getPercentPoint(testValueRelPercentileValue).doubleValue();
+                        relWindowStatCalc = new SamplingStatCalculator();
+                    }
+                    if (previousWindowPercentile > 0) {
                         double growthPct = (currentPct - previousWindowPercentile) * 100.0 / previousWindowPercentile;
                         if (growthPct >= testValueRelThresholdPct) {
                             log.info("P" + (int)(testValueRelPercentileValue * 100) + " grew " + String.format("%.1f", growthPct) + "% (" + (int)previousWindowPercentile + " ms -> " + (int)currentPct + " ms) over " + testValueRelWindowSecs + "s window, threshold " + testValueRelThresholdPct + "%. Auto-shutdown test...");
@@ -183,20 +191,19 @@ public class AutoStop
                         }
                     }
                     previousWindowPercentile = currentPct;
-                    relWindowStatCalc = new SamplingStatCalculator();
                     relWindowStart = sec;
                 }
             }
 
             // Per-window error count ceiling check
             if (testValueErrorCount > 0 && testValueErrorCountSec > 0) {
-                if (errorCountInWindow > testValueErrorCount) {
-                    log.info("Error count more than " + getErrorCount() + " for " + getErrorCountSecs() + "s. Auto-shutdown test...");
-                    System.out.println("AutoStop - Error count more than " + getErrorCount() + " for " + getErrorCountSecs() + "s. Auto-shutdown test...");
+                if (errorCountInWindow.get() > testValueErrorCount) {
+                    log.info("Error count more than " + getErrorCount() + " within " + getErrorCountSecs() + "s. Auto-shutdown test...");
+                    System.out.println("AutoStop - Error count more than " + getErrorCount() + " within " + getErrorCountSecs() + "s. Auto-shutdown test...");
                     stopTest();
                 }
                 if (sec - errorCountWindowStart >= testValueErrorCountSec) {
-                    errorCountInWindow = 0;
+                    errorCountInWindow.set(0);
                     errorCountWindowStart = sec;
                 }
             }
@@ -251,12 +258,14 @@ public class AutoStop
         testValueRelPercentileValue = getRelPercentileValueAsFloat();
         testValueRelWindowSecs = getRelWindowSecsAsInt();
         testValueRelThresholdPct = getRelThresholdPctAsFloat();
-        relWindowStatCalc = new SamplingStatCalculator();
+        synchronized (relWindowLock) {
+            relWindowStatCalc = new SamplingStatCalculator();
+        }
         relWindowStart = System.currentTimeMillis() / 1000;
         previousWindowPercentile = -1;
         testValueErrorCount = getErrorCountAsInt();
         testValueErrorCountSec = getErrorCountSecsAsInt();
-        errorCountInWindow = 0;
+        errorCountInWindow.set(0);
         errorCountWindowStart = System.currentTimeMillis() / 1000;
     }
 
@@ -397,7 +406,7 @@ public class AutoStop
     private float getPercentileValueAsFloat() {
         String val = getPercentileValue();
         if (val == null || val.trim().isEmpty()) {
-            return 0;
+            return 1;
         }
         float res = 0;
         try {
@@ -406,7 +415,7 @@ public class AutoStop
             log.error("Wrong Percentile Value: " + val, e);
             setPercentileValue("1");
         }
-        return res > 0 ? res : 0;
+        return res > 0 ? res : 1;
     }
 
     private int getResponseTimeAsInt() {
--- a/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
+++ b/plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
@@ -49,6 +49,13 @@ public class AutoStop
     private final static String PERCENTILE_RESPONSE_TIME_SECS ="percentile_response_time_secs";
     private final static String PERCENTILE_VALUE = "percentile_value";
     private final static String CUSTOM_VALIDATION_DURATION = "custom_validation_duration";
+    // New: relative window-to-window percentile degradation
+    private final static String REL_PERCENTILE_VALUE = "rel_percentile_value";
+    private final static String REL_PERCENTILE_WINDOW_SECS = "rel_percentile_window_secs";
+    private final static String REL_PERCENTILE_THRESHOLD_PCT = "rel_percentile_threshold_pct";
+    // New: per-window error count ceiling
+    private final static String ERROR_COUNT = "error_count";
+    private final static String ERROR_COUNT_SECS = "error_count_length";
     private long curSec = 0L;
     private GraphPanelChartAverageElement avgRespTime = new GraphPanelChartAverageElement();
     private GraphPanelChartAverageElement avgRespLatency = new GraphPanelChartAverageElement();
@@ -76,6 +83,17 @@ public class AutoStop
     private int testValueCustomSec = 180;
     private boolean skipIter = false;
 
+    private SamplingStatCalculator relWindowStatCalc = new SamplingStatCalculator();
+    private long relWindowStart = 0;
+    private double previousWindowPercentile = -1;
+    private float testValueRelPercentileValue = 0;
+    private int testValueRelWindowSecs = 0;
+    private float testValueRelThresholdPct = 0;
+    private int errorCountInWindow = 0;
+    private long errorCountWindowStart = 0;
+    private int testValueErrorCount = 0;
+    private int testValueErrorCountSec = 0;
+
 
     public AutoStop() {
         super();
@@ -84,6 +102,14 @@ public class AutoStop
     @Override
     public void sampleOccurred(SampleEvent se) {
         long sec = System.currentTimeMillis() / 1000;
+
+        if (testValueRelWindowSecs > 0) {
+            relWindowStatCalc.addSample(se.getResult());
+        }
+
+        if (testValueErrorCount > 0 && !se.getResult().isSuccessful()) {
+            errorCountInWindow++;
+        }
 
         if (curSec != sec) {
             if (testValueRespTime > 0) {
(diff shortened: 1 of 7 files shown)


plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -85,6 +103,14 @@ public AutoStop() {
     public void sampleOccurred(SampleEvent se) {
         long sec = System.currentTimeMillis() / 1000;
 
+        if (testValueRelWindowSecs > 0) {
+            relWindowStatCalc.addSample(se.getResult());
```
undera: **Blocker: unsynchronized accumulator on the per-sample path.**

`AutoStop` is `NoThreadClone`, so one instance is shared and `sampleOccurred` runs concurrently on every JMeter thread. `SamplingStatCalculator.addSample(SampleResult)` is not synchronized (verified in ApacheJMeter_core 2.13 and 5.6.3 — only `clear()` is), and it delegates to `StatCalculator.addValue`, also unsynchronized and backed by a plain `java.util.TreeMap`. So this is concurrent unguarded mutation of a red-black tree, with `getPercentPoint` reading it unguarded at the window boundary. Under load that can corrupt the map, throw `ConcurrentModificationException`, or return wrong percentiles.

This is a different exposure from the existing `statCalc.addSample(sr)` call, which sits inside `if (curSec != sec)` and so fires about once per second. Accumulating on every sample from every thread makes the race the normal case rather than a rare one.

Please synchronize access to the accumulator, or accumulate per-thread and merge at the window boundary.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -144,6 +170,37 @@ public void sampleOccurred(SampleEvent se) {
                 }
             }
 
+            // Relative window-to-window percentile degradation check
+            if (testValueRelWindowSecs > 0 && testValueRelThresholdPct > 0) {
```
undera: **Blocker: a blank percentile rank still runs the check, against P0.**

The guard omits `testValueRelPercentileValue`. With the rank field blank, `getRelPercentileValueAsFloat()` returns `0`, so `getPercentPoint(0)` returns the window *minimum* — the criterion then compares minimum response times window over window, and will stop the test.

Reproduced on this branch:

```java
AutoStop a = new AutoStop();
// rel_percentile_value deliberately not set
a.setRelWindowSecs("1");
a.setRelThresholdPct("10");
a.testStarted();
for (int i = 0; i < 5; i++) a.sampleOccurred(ev(100));
Thread.sleep(1100);
for (int i = 0; i < 5; i++) a.sampleOccurred(ev(100));
Thread.sleep(1100);
for (int i = 0; i < 5; i++) a.sampleOccurred(ev(900));
Thread.sleep(1100);
a.sampleOccurred(ev(900));
// System.getProperty("auto_stopped") == "true"
```

Adding `testValueRelPercentileValue > 0` to the guard fixes it.

For the record the intended path does work: rank 90, threshold 50%, 100ms -> 1000ms stops the test as designed.

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
```diff
@@ -245,4 +247,184 @@ public void testGetPercentileValue() {
...
+
+    @Test
+    public void testSetGetRelThresholdPct() {
+        System.out.println("setGetRelThresholdPct");
+        AutoStop instance = new AutoStop();
+        instance.setRelThresholdPct("20");
+        assertEquals("20", instance.getRelThresholdPct());
+    }
+
+    /** Feature is a no-op (no exception) when percentile rank is not configured. */
```
undera: This comment states the opposite of the actual behaviour, and the test asserts nothing, so nothing catches it — see my note on the guard in `AutoStop.sampleOccurred`. This exact configuration (window and threshold set, rank blank) does stop the test, via a P0 comparison.

Please add an assertion on `auto_stopped` here; it is the test that would have caught the guard bug.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -271,115 +350,226 @@ String getErrorRateSecs() {
...
         try {
-            res = Float.parseFloat(getPercentileValue()) / 100;
+            res = Float.parseFloat(val.trim()) / 100;
         } catch (NumberFormatException e) {
-            log.error("Wrong Percentile Value: " + getPercentileValue(), e);
+            log.error("Wrong Percentile Value: " + val, e);
             setPercentileValue("1");
         }
-        return res > 0 ? res : 1;
+        return res > 0 ? res : 0;
```
undera: **Undeclared change to the existing percentile criterion.**

This fallback changed from `res > 0 ? res : 1` to `: 0`. For a plan with `percentile_response_time` set and `percentile_value` blank, that converts a P100 check into a P0 check — a criterion that used to fire now never fires.

Reaching that state needs a hand-edited or partially-written JMX, so I don't think it is common in the field, but it is a silent behaviour change to an existing shutdown criterion and it isn't mentioned in the PR description. Please restore the previous fallback, or call the change out explicitly with the reasoning.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -85,6 +103,14 @@ public AutoStop() {
     public void sampleOccurred(SampleEvent se) {
         long sec = System.currentTimeMillis() / 1000;
 
+        if (testValueRelWindowSecs > 0) {
+            relWindowStatCalc.addSample(se.getResult());
+        }
+
+        if (testValueErrorCount > 0 && !se.getResult().isSuccessful()) {
+            errorCountInWindow++;
```
undera: Two issues on this counter:

1. `errorCountInWindow++` is a non-atomic read-modify-write on a field shared across all threads (`NoThreadClone`), so it undercounts under load and delays the stop.
2. This guard is `testValueErrorCount > 0`, but the evaluation guard below is `testValueErrorCount > 0 && testValueErrorCountSec > 0`. When the window is left unset the counter increments without bound and is never read. Please make the two guards agree.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -144,6 +170,37 @@ public void sampleOccurred(SampleEvent se) {
                 }
             }
 
+            // Relative window-to-window percentile degradation check
+            if (testValueRelWindowSecs > 0 && testValueRelThresholdPct > 0) {
+                if (sec - relWindowStart >= testValueRelWindowSecs) {
+                    double currentPct = relWindowStatCalc.getPercentPoint(testValueRelPercentileValue).doubleValue();
+                    if (previousWindowPercentile >= 0 && previousWindowPercentile > 0) {
```
undera: `previousWindowPercentile >= 0 && previousWindowPercentile > 0` — the first term is redundant, `> 0` already implies it.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
```diff
@@ -144,6 +170,37 @@ public void sampleOccurred(SampleEvent se) {
...
+                        }
+                    }
+                    previousWindowPercentile = currentPct;
+                    relWindowStatCalc = new SamplingStatCalculator();
+                    relWindowStart = sec;
+                }
+            }
+
+            // Per-window error count ceiling check
+            if (testValueErrorCount > 0 && testValueErrorCountSec > 0) {
```
undera, line 199: The semantics here are a tumbling window ("more than N errors *within* M seconds"), whereas the five criteria above all mean "sustained *for* M seconds". That's a reasonable choice, but please make it explicit in the log message, which currently reads "Error count more than N for Ms" and so implies the sustained meaning.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/JAutoStopPanel.java
```diff
@@ -20,45 +30,325 @@ public JAutoStopPanel() {
...
-        jTextFieldErrorSec.setText(testElement.getErrorRateSecs());
-        jTextFieldRespLatency.setText(testElement.getResponseLatency());
-        jTextFieldRespLatencySec.setText(testElement.getResponseLatencySecs());
-        jTextFieldPercentileRespTime.setText(testElement.getPercentileResponseTime());
-        jTextFieldPercentileRespTimeSec.setText(testElement.getPercentileResponseTimeSecs());
-        jTextFieldPercentileValue.setText(testElement.getPercentileValue());
+        boolean hasRespTime = isConfigured(testElement.getResponseTime());
+        jCheckBoxRespTime.setSelected(hasRespTime);
+        jTextFieldRespTime.setText(hasRespTime ? testElement.getResponseTime() : "10000");
+        jTextFieldRespTimeSec.setText(isConfigured(testElement.getResponseTimeSecs()) ? testElement.getResponseTimeSecs() : "10");
```
undera: `configure()` now substitutes hardcoded defaults into the text fields for values the element does not have. Since `modifyTestElement` writes the field contents straight back, opening an existing plan and saving it can persist values the user never chose — a blank seconds field becomes `"10"` here, where the runtime default was `1`.

Prefer leaving unset fields empty on load, so loading and saving a plan is a no-op.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/JAutoStopPanel.java
```diff
@@ -20,45 +30,325 @@ public JAutoStopPanel() {
...
+        jPanel5.add(jCheckBoxPercentile, 0);
+
+        // Adjust OR label indentation
+        adjustOrLabelInsets(jLabel8);
+        adjustOrLabelInsets(jLabel9);
+        adjustOrLabelInsets(jLabel13);
+
+        // Remove unused placeholder components to eliminate extra blank rows between criteria
+        remove(jLabel20);
+        remove(jPanel6);
```
undera, line 227: This drops the existing custom-validation row from the panel, while `configure()` and `modifyTestElement` still read and write `custom_validation_duration` — so the property survives but becomes uneditable.

The row is an unlabelled, half-finished control on master, so retiring it is defensible. If that is the intent, please remove the property handling along with it and say so in the description; otherwise leave the row in place.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/JAutoStopPanel.java
```diff
@@ -20,45 +30,325 @@ public JAutoStopPanel() {
...
+        jPanelErrorCount.add(jLabelBulletErrorCount);
+
+        jPanelErrorCount.add(new JLabel("Error Count is greater than"));
+
+        jTextFieldErrorCount = new JTextField(5);
+        jTextFieldErrorCount.setHorizontalAlignment(JTextField.RIGHT);
+        jTextFieldErrorCount.setToolTipText("Maximum allowed error count within the duration.");
+        jPanelErrorCount.add(jTextFieldErrorCount);
+
+        jPanelErrorCount.add(new JLabel("for"));
```
undera: Reads as "Error Count is greater than 10 **for** 10 seconds", which matches the "sustained for" wording of the five rows above but not the actual tumbling-window behaviour. Suggest "within" here.

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
```diff
@@ -245,4 +247,184 @@ public void testGetPercentileValue() {
...
+        instance.testStarted();
+        SampleEvent event = new SampleEvent(resultWithTime(200), "");
+        instance.sampleOccurred(event);
+        synchronized (this) { wait(1100); }
+        instance.sampleOccurred(event);
+    }
+
+    /** Samples accumulate across two windows without error. */
+    @Test
+    public void testRelativeWindowAccumulatesAcrossTwoWindows() throws InterruptedException {
```
undera: This test asserts nothing, and it sets up the one case that matters most: 100ms -> 500ms against a 50% threshold, i.e. a genuine breach of the new criterion. As written, the percentile stop path has no coverage at all.

Please assert `auto_stopped` is set here. Same for `testErrorCountDisabledWhenNotConfigured` below.

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
```diff
@@ -245,4 +247,184 @@ public void testGetPercentileValue() {
...
+    public void testSetGetErrorCountSecs() {
+        System.out.println("setGetErrorCountSecs");
+        AutoStop instance = new AutoStop();
+        instance.setErrorCountSecs("10");
+        assertEquals("10", instance.getErrorCountSecs());
+    }
+
+    /** Feature is disabled when error count is not configured. */
+    @Test
+    public void testErrorCountDisabledWhenNotConfigured() throws InterruptedException {
```
undera, line 429: Also asserts nothing. Please assert that `auto_stopped` is *not* set, so the disabled path is actually pinned down.

Separately: the `wait(1100)` calls across the new tests take `AutoStopTest` from about 8s to 65s. Shortening the windows would keep a minute of sleeping out of CI.

undera (changes requested):
Thanks — this targets the right place, and the build is green here too (40 tests, 0 failures). I confirmed the intended percentile path works: rank 90, threshold 50%, 100ms -> 1000ms stops the test as designed.

Two blockers before this can go in, both left as inline comments:

1. `relWindowStatCalc.addSample()` runs on the per-sample path without synchronization, in a `NoThreadClone` listener shared across all threads.
2. A blank percentile rank does not disable the new criterion — it runs it against P0 and can stop the test. The test that covers that configuration asserts nothing and its comment claims the opposite.

The rest of the inline notes are smaller: three assert-free tests, an undeclared change to the existing percentile fallback, the removed custom-validation row, and some wording and guard mismatches.

One doc item not tied to a line: `site/dat/wiki/AutoStop.md` still says "Currently 3 criteria available". It was already behind by one (percentile), and this makes six — good moment to bring it up to date, including the window semantics for both new criteria.

Version bump and the marketplace `versions` entry I will handle at release time; no need to touch those here.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
hilalsidhic: Fixed. Added a dedicated relWindowLock monitor object. Both the per-sample addSample() call and the window-boundary getPercentPoint() + reset are now guarded by synchronized(relWindowLock). The testStarted() reset is also synchronized. SamplingStatCalculator's unsynchronized TreeMap is never accessed concurrently anymore.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
hilalsidhic: Fixed. Both the per-sample accumulation guard and the window-boundary evaluation guard now require testValueRelPercentileValue > 0 in addition to the window and threshold checks. A blank or zero rank completely skips the criterion on every code path.

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
hilalsidhic: Fixed. Rewrote the test as testRelativeWindowDisabledWhenNotConfigured, It now asserts assertNull(System.getProperty("auto_stopped"))

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
hilalsidhic: Restored. getPercentileValueAsFloat() now returns 1 when the value is null/empty, and res > 0 ? res : 1 on a parse error — exactly matching the original behaviour before our changes. The new getRelPercentileValueAsFloat() method correctly returns 0 when unset, so the two criteria behave independently and correctly.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
hilalsidhic: errorCountInWindow is now AtomicInteger, with incrementAndGet(), get(), and set(0) throughout.
The increment guard is now testValueErrorCount > 0 && testValueErrorCountSec > 0, matching the evaluation guard exactly. The counter will not increment when the window duration is unset.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/AutoStop.java
hilalsidhic: Fixed. Both the log.info and System.out.println for the error count shutdown now read "within": "Error count more than N within Ms. Auto-shutdown test...".

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/JAutoStopPanel.java
hilalsidhic: Fixed. configure() now sets each text field directly from the element's property value, with an empty string fallback for null. No synthetic defaults are injected. initFields() (used only when creating a brand-new element) still sets helpful placeholder values, which is the correct place for them. A testConfigureUnsetFieldsRoundTrip test has been added to JAutoStopPanelTest to pin this behaviour.

plugins/autostop/src/main/java/kg/apc/jmeter/reporters/JAutoStopPanel.java
hilalsidhic: Fixed. Both configure() and modifyTestElement() no longer call getCustomValidationDuration() or setCustomValidationDuration(). The panel continues to remove jPanel6 visually (to eliminate the blank row in the GUI). The property getter/setter on AutoStop are preserved for backward compatibility with any existing JMX files that may have the value saved, but the UI no longer reads from or writes to it.

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
hilalsidhic: Fixed. Renamed and rewrote as testRelativeWindowBreachStopsTest. It uses 100ms → 500ms across a 1s window against a 50% threshold and asserts assertEquals("true", System.getProperty("auto_stopped")), pinning down the stop path.

undera (approved):

plugins/autostop/src/test/java/kg/apc/jmeter/reporters/AutoStopTest.java
hilalsidhic: Fixed. Added assertNull("Unconfigured error count must not stop test", System.getProperty("auto_stopped")). The test also now calls System.clearProperty("auto_stopped") at the start to prevent cross-test pollution.

All synchronized(this) { wait(1100); } monitor-wait patterns have been replaced with standard Thread.sleep(1100). All existing scenarios were preserved (and new tests added for concurrency and round-trip configuration), with the full test suite completing in ~28 seconds. Since JMeter's window epoch evaluates in whole seconds (System.currentTimeMillis() / 1000), the 1.1s sleep represents the minimum reliable margin across window transitions.

PR #888, opened 2026-09-16
Add parallel controller release 0.13

diego-ferrand:


undera:
The `jmeter-parallel-0.13.jar` on Maven Central bundles ~1,100 `org/apache/jmeter/**` classes from JMeter 3.1 and some other libs, which 0.12 did not. Is that intentional?

diego-ferrand:
Not intentional. Shading was added in 0.13, but it bundled `ApacheJMeter_http 3.1` because we forgot to include provided in the POM.

I need to review whether shading is even necessary, since jmeter-plugins-cmn-jmeter is our only actual dependency. We were initially looking at shading to make it easier for users who don't use the Plugin Manager to manually install the plugin by simply copying over an uber JAR.

I'll close this PR and open a new one once the hotfix release is ready.

